PR #13, opened 2015-07-21
Bugfix/navigate parameter type

bc3tech:
I think this should stay `object` since that's what the UWP API `Frame.Navigate()` supports...


JerryNixon:
No. String is the only reliable parameter.

To enable frame state serialization using GetNavigationState, you must use only basic types for the navigation parameter, such as string, char, numeric, and GUID types. Otherwise GetNavigationState will throw an exception when the app suspends. The parameter can have other types if you do not use GetNavigationState.

https://msdn.microsoft.com/en-us/library/windows/apps/hh702394.aspx

That being said, we should consider an alternative for simple scenarios.

PR #14, opened 2015-07-23
Bugfix/back button crash

bc3tech:


bc3tech:
@JerryNixon found the root cause. The `args.handled=true` _every time_ `BackRequested` fires is the error. in the case where we're at top level, we won't actually end up handling anything, so we should let the OS take care of it... right?

JerryNixon:
Ah, yeah. So consider that it is common for an app to have more than one Frame. That means more than one NavService. The library already supports this. What do you suggest?

bc3tech:
Multiple frames = multi window app on "non-mobile" device family, right?

tibitoth:
@bc3tech Not only. In Jerry's SplitView pattern there are multiply Frames in phones too.

bc3tech:
haven't looked at this. is it possible on, say, a phone?

tibitoth:
Oh sorry, there aren't two Frames in Jerrys SplitView samle but it is possible to use one Frame outside the SplitView and one inside the SplitView. http://blog.jerrynixon.com/2015/04/implementing-hamburger-button-with.html

bc3tech:
Here's my (theoretical) thinking:
`BackPressed` should only set `handled = true` if it does actually handle something.
That being said, it's the developer's job for handling the behavior if & when it makes sense to do so. So, for the case of multiple frames I had better, as a developer, handle what the back button means for my frame (as in the split view, to collapse it) and therefore set `handled = true` when I do.
For multi-window applications, I think the "fix" shown in this PR is still valid. On desktop, multi window app have their own Back (when possible) button and a **close** button for when the user can no longer go back. For mobile apps, we don't really have the concept of multi-window, so hitting back in a new window until you're at the top of the stack should (and will, I believe) close that secondary window and take you back to the main app.

Admittedly I've not worked with multi-frame/window UWPs yet.

PR #17, opened 2015-08-06
Changed RootNamespace to Template10

mattiascibien:
I changed this so, first of all, ReSharper does not ask to change the references to Template10Library and then every new file created will have the Template10 namespace.


JerryNixon:
You inspired us to change this, but we did in a separate branch that has merged. Thank you for your contribution to Template 10. I'm looking forward to releasing it in the next few days.

mattiascibien:
Nice to hear it. ;)

Issue #48, opened 2015-08-27
Parameter object passed to NavigationService.Navigate is cast as string


joshberry:
The NavigationService is casting the parameter argument of the Navigate method to a string before I receive it in the OnNavigatedTo method of my receiving page's view model . Is this by design? I suspect it's a bug or else the parameter argument should just be a string in the first place.

For example, if I call the Navigate method like so:

``` csharp
NavigationService.Navigate(typeof(Views.MyPage), myCustomObject);
```

When I receive myCustomObject in the view model it's a string:

``` csharp
public override void OnNavigatedTo(object parameter, NavigationMode mode, IDictionary<string, object> state)
{
    // the parameter with myCustomObject is now a string
}
```

dmit25:
I assume it just should be serializable. It would be nice to have automatic serialization to json, it would allow to save context in file or smth so it could be used later when restorig application state.

JerryNixon:
I will check this out. But @dmit25 automatic serialization to JSON is one thing, automatic deserialization from JSON is prone to failure for generic implementations and more. The behavior should be that it is not turned into a string. That's a genuine bug if it is happening.

joshberry:
Take a look at my pull request and you'll see exactly where it's happening.

darenm:
The problem has been confirmed and corrected. Please get latest. 

tldr; The issue was in the NavigatedEventArgs - a ToString() had been left in from an earlier version where we experimented with passing strings only.

PR #49, opened 2015-08-27
Allow non-string types to be passed as navigation parameter

joshberry:
Resolving issue #48


JerryNixon:
We've already corrected this directly. Thank you.

PR #50, opened 2015-08-27
initial testcases, settings

kopfkraft:
initial setup of a test library. settings changed to improve testability.


JerryNixon:
Please reduce your pull request to smaller pieces. Settings first, please. Also change your implementation so it does not instantiate both Local and Roaming when only Local is used. Second, please keep the convention of SettingService. Also please smooth out the API so it is either Setting or Settings. Also, good work. This is very good work. I am closing this pull request, looking forward to your next.

PR #56, opened 2015-08-29
Hamburgermenu enhancement

dg2k:
HamburgerMenu Dark and Light themes are incomplete; for instance, for
RequestedTheme=Light | Dark definitions in
Minimal\Styles\Custom.xaml to work, HamburgerMenu control needs some
enhancement. This modified version works for both Dark and Light themes
by setting the RequestedTheme attribute in App.xaml (demo with Minimal
project). Of course user can define in Styles\Custom.xaml as suited but for demo some Light styles are included.

PR #64, opened 2015-09-04
Re-submission of #56 (Take 3 ;-)

dg2k:
Re-submission of #56 (Take 3 ;-)


JerryNixon:
This submission can no longer be merged without some work.
I like this. Sorry for take 4.

PR #70, opened 2015-09-09
Modifications in HamburgerMenu

abubberman:
- Removed the margin in PaneContent. The pane and hamburger button are
  now better aligned.
- Added a itemscontrol to the header section. This enables custom
  controls in the header, like a search bar

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -168,7 +168,10 @@
     </UserControl.Resources>
 
     <Grid x:Name="RootGrid">
-
+        <Grid.RowDefinitions>
+            <RowDefinition Height="48"/>
+            <RowDefinition/>
+        </Grid.RowDefinitions>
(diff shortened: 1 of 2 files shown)


jmbezeau:
Could you provide an example of xaml code to implement a search bar?

Thanks

abubberman:
See screenshot for implementation. Code sample did not show up correctly in preview.. so i used a screenshot instead. It's also important to add a top margin in the container control on your pages. Otherwise the content of the page will be shown on top of the searchbar. In my app, i used the following: Margin="0,48,0,0".

image

jmbezeau:
Thank you! I was hoping to get the searchbox in the MainPage.xaml file, is there a way to have the Searchbox visible on only some pages?

abubberman:
Well, i guess you could add a setting which triggers the visibility of the searchbox. Also, you could remove the margin in the pages that do not have to show the searchbox. Then the content of the page will be on top of the searchbox. 

As far as I know, there is no Property available to toggle the HeaderContent.

PR #71, opened 2015-09-09
Fixed minor bug not setting/showing entered the BusyText correctly

ghost:
We found a minor issue that the entered busy text is not shown in the sample. So we fixed it.


teamneusta:
Duplicate to #66

PR #72, opened 2015-09-09
Make Template10 assemblies strong-named.

ghost:
We find it useful to add a strong name key to Template10 nuget packages. Not to increase security or something like this, but to make it useable for projects which must be strong-named. With this pull request, we added a strong name to Template10 library and services assemblies.


JerryNixon:
Great addition. Could you re-sub this request after syncing?

PR #75, opened 2015-09-10
Added fix for issue #72 and solved reference issues

callummoffat:
This pull request adds the enhancement I detailed in Issue #72 (it uses the system accent color for the title bar by default). It also fixes an issue I've found, whereby a missing dependency on the Visual C++ 2015 Runtime for UWP wreaks project-wide havoc. I fixed it by (_gasp_) adding the missing dependency. Don't know how that got missed in the first place, but it's fixed now!


callummoffat:
I've added a few changes since my original pull request. This pull request also uses the accent by default for the backgrounds of PageHeader and Hamburger menu. Feel free to test the heck out of this!

PR #79, opened 2015-09-10
Title Bar Enhancements

callummoffat:
Continuing with my campaign to improve Template 10's UI, especially in the area of the colors used by default, this pull request adds the system accent color as the default title bar color. It does this in the project templates OnStartAsync(), rather than in BootStrapper.cs within the library, which makes it easier to change the implementation for newbies. I've also added some documentation about this into the code, to make it easier for newbie programmers.

There are, however, a few items with this particular pull request that still need to be implemented. I haven't yet figured out how to create hover/click/inactive background colors that are usable with all color combinations. At this stage, they just use the defaults, because I'm not quite sure how to implement that yet.

### DIFF
--- a/Samples/Tiles/Tiles.csproj
+++ b/Samples/Tiles/Tiles.csproj
@@ -140,6 +140,9 @@
     <SDKReference Include="BehaviorsXamlSDKManaged, Version=12.0">
       <Name>Behaviors SDK %28XAML%29</Name>
     </SDKReference>
+    <SDKReference Include="Microsoft.VCLibs, Version=14.0">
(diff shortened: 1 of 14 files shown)


callummoffat:
I have now fixed the issues regarding the title bar color (not being able to select a good-looking default color for hover/click/inactive backgrounds). I did it by adding a new converter. Currently, this functionality is enabled through an XAML converter in the namespace of each project. I intend to submit an update integrating the converter into the main Template10Library as soon as the library comes out of a state of flux, when it will be easier to add new features.

PR #82, opened 2015-09-11
Hide the Hamburger Menu

ghost:
Add property to hide the menu completely, useful to show full-screen content like a video player.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -136,11 +136,11 @@
                                             </VisualState.Setters>
                                         </VisualState>
                                         <VisualState x:Name="Disabled"/>
-										<VisualState x:Name="Hidden">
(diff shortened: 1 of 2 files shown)


JerryNixon:
Great idea.

JerryNixon:
I have implemented HamburgerMenu.IsFullScreen.

SBortz:
You´re awesome :)
The new title bar broke our solution to this.

JerryNixon:
Thanks. How?

Best regards,
Jerry Nixon

Sent from my Windows Phone.
Please forgive typos and brevity.

SBortz:
Not important. We were following a different approach with another parent Grid with 2 VisualStates. Your solution looks great, tomorrow i will have a closer look to it.

teamneusta:
Jerry, I come back to this issue - I get a COM exception when setting IsFullScreen to true. I was able to fix this by adding a missing =null assignment, but the IsFullScreen property has no effect at all. I will file a new issue on this.

PR #102, opened 2015-09-16
Modified HeaderItems ItemsControl

abubberman:
Modified the HeaderItems ItemControl. Changed the ItemsPanelTemplate to use a StackPanel with a horizontal orientation and a righttoleft flowdirection


JerryNixon:
Double check against current code

abubberman:
Ah.. did'nt see the latest version. Will double check the next time

JerryNixon:
Please see

> https://github.com/Windows-XAML/Template10/wiki/Docs-%7C-PageHeader
> Will need to be updated, right?

PR #142, opened 2015-09-25
Adding the ability to associate a key to a page to allow navigation without specifying the page type.

marcominerva:
This is useful for example when using NavigationService from a ViewModel in an MVVM scenario: in this case, the ViewModel shouldn't know anything about pages.

### DIFF
--- a/Template10 (Library)/Services/NavigationService/INavigationService.cs
+++ b/Template10 (Library)/Services/NavigationService/INavigationService.cs
@@ -22,8 +22,11 @@ namespace Template10.Services.NavigationService
         void ClearHistory();
         void GoBack();
         void GoForward();
+        void Register(string key, Type page);
         bool Navigate(Type page, object parameter = null, NavigationTransitionInfo infoOverride = null);
+        bool Navigate(string pageKey, object parameter = null, NavigationTransitionInfo infoOverride = null);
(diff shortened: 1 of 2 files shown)


JerryNixon:
Okay, so this is a great idea. But this is not the right approach. At least not for T10. I took the work you did and refactored it into two methods. The first is Bootstrapper.PageKeys<T>() and the second is NavigationService.Navigate<T>(). Both are written and committed and I also included a new PageKeys sample project to demonstrate it. What I need from you, Marco, is to complete the readme.md in that sample project to explain to people how to properly use it. I added some starter text to help you. Since you are clearly very familiar with this approach, I am not concerned. One more thing, the T10 approach is an Enum approach, not a string approach. The Key of the View MUST be an Enum. When you look at the implementation, you will immediately understand and probably fall in love with it.

marcominerva:
Thank you, I have just made a new pull request following you suggestions. And yes, I love the Enum approach :-)

ryanzzff:
Sorry, I didn't get the idea. Why would this approach would be better than the previous one?

You init the page keys at OnInitializeAsync(),
the ViewModels will not know the PageType, yes, but seems the Enum and the PageType is one-to-one mapping,
if the ViewModels have to know which Enum to navigate, I see no difference between it and navigating by PageType directly.

On the other hand, it seems more complicated, as when I add a new Page, I have to add a new Enum, and add a mapping at OnInitializeAsync().
Or I missed something? thanks.

PR #146, opened 2015-09-25
Updated readme.md to explain how to use PageKeys for navigation

marcominerva:
This approach allows navigation by key instead of page types (This is useful for example when using NavigationService from a View Model in an MVVM scenario).


JerryNixon:
Nice work.

marcominerva:
Thank you :-)

PR #148, opened 2015-09-26
Add ClearCache in NavigationService

pekspro:


JerryNixon:
Sure. Why not? When would you use this?

pekspro:
Thanks for merging :-). I’ve used to completely reset some views. For instance select the first pivot item, reset scroll bar in lists etc. Maybe there a more clever ways to do it, but I found this way quite handy :-).

Issue #155, opened 2015-09-28
HamburgerMenu Display Mode


dg2k:
Don't know if I'm over-looking the obvious -- what's an easy way to find out the DisplayMode (Overlay, CompactOverlay, Inline)  for the HM in the Shell.xaml codebehind? I needed this 'cos I have a notification status bar at the bottom of screen that needs its left margin set conditionally (48 in CompactOverlay mode so as not to cover the Settings icon, 0 otherwise as in Mobile).

If this is not currently possible, do you see the need for this feature as I obviously do?

youtoofan:
Why don't you use 'VisualStateGroups' ?

dg2k:
Can you Q.E.D your statement please?

Although I can see the defined visual states quite easily, I cannot see an object that gives me the CurrentState of the VisualManager.

JerryNixon:
Every VisualState is a child of a VisualStateGroup. The VisualStateGroup is an object with properties, including CurrentVisualState. Check it out. That being said, I think it makes sense to have a public, readonly property to let the developer test what the current display mode is. Let me think about it.

JerryNixon:
Added the property. It is readonly. It is also a dependency property, so you can bind to it and listen to it.

dg2k:
Thanks a lot Jerry.

PR #156, opened 2015-09-28
Correcting System.InvalidCastException when accessing the PaneWidth p…

dg2k:
While for a lookout in HamburgerMenu code relating to my posting under #155 I came across a small bug (call it a bugette ;-) that is self-correcting and therefore hard to detect.

Attempting to access the PaneWidth property of HamburgerMenu before navigating to another page (ShellSplitView.IsPaneOpen == false ) throws an exception of System.InvalidCastException("Specified cast is not valid"). (This may have an impact in HamburgerMenu.xaml Binding but haven't checked thoroughly;  interestingly, in previous versions this was FURTHER disguised by a FallbackValue for the binding).

If navigated from the landing page, MainPage.xaml, (with ShellSplitView.IsPaneOpen == true) the problem corrects itself as PaneWidth is explicitly initialized by the NavigationService property setter, hence hiding the problem.

As it happens, this is a single letter correction (you can't beat this as the smallest pull request!): the default value of 220 passed on to the PaneWidth DependencyProperty should be set as 220d with the double suffix d. This is a well known oversight as well explained here: http://reedcopsey.com/2009/10/23/wpf-common-dependency-property-exception/ as to how the missing suffix d leads to System.InvalidCastException.

Surely some feedback for the compiler design team as coding for DependencyProperty has unambiguous declaration with typeof(double) which the compiler can easily deduce (or at least issue a warning).

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
@@ -407,7 +407,7 @@ namespace Template10.Controls
         }
         public static readonly DependencyProperty PaneWidthProperty =
             DependencyProperty.Register("PaneWidth", typeof(double),
-                typeof(HamburgerMenu), new PropertyMetadata(220));
+                typeof(HamburgerMenu), new PropertyMetadata(220d));
 
         public UIElement HeaderContent
         {


JerryNixon:
Brilliant work.

PR #158, opened 2015-09-28
fixes a bug that resets the IsEnabled property of HamburgerButtonInfos on Navigation

MichaelPetrinolis:


JerryNixon:
Several other things had to be changed to make this work. The login sample is the right place to test if it works. I have implemented your login in the main branch but cannot merge this pull request.

PR #171, opened 2015-10-01
CompactInline-CompactOverlay toggle menu mode

dg2k:
In wide display mode, sometimes it may be desirable to collapse the CompactInline menu to the CompactOverlay mode. This pull request implements a toggle effect (you enable it only when you need it in wide display mode). How about we call it, ... CompactInline - CompactOverlay  or CICO (ouch!! does that remind you of 'model view view model' ? ;-).Anyways, see what you think.

In addition to the main implementation in the HamburgerMenu control (VisualStates and DP's), a setting via Settings Service is also implemented and demonstrated via Minimal template project.
Hope you'll like it..

[Overlay]  [CompactOverlay]  [Compactnline]  [CompactInline-CompactOverlay]  [Inline] ... [Weird, huh!]

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -178,7 +178,7 @@
     <Grid x:Name="RootGrid">
 
         <VisualStateManager.VisualStateGroups>
-            <VisualStateGroup x:Name="VisualStateGroup">
+            <VisualStateGroup x:Name="VisualStateGroup" CurrentStateChanged="OnVisualStateGroupChanged">
 
                 <VisualStateGroup.Transitions>
                     <VisualTransition GeneratedDuration="0" To="VisualStateWide">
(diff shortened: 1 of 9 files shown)


JerryNixon:
I don't understand. Do you have a screenshot?

dg2k:
Easy to explain.

CompactOverlay kicks in at NormalMinWidth of 521.
CompactInline kicks in at WideMinWidth of 1200.

Toggle0 - every thing as normal; nothing changes.
Toggle1 - CompactOverlay all the way from 521 and over. Why? We save the screen real-estate taken up by the open pane in CompactInline mode. Some apps just look nice this way. It is simply an option, not an imposition. The functionality doesn't even exist for screen size < NormalMinWidth  as the trigger for the visual state doesn't exist anyway.

Setting implementation in Minimal template project is shown as you did for Dark/Light Theme.

JerryNixon:
I think this already does what you are asking for, no?

``` XAML
<HamburgerMenu.VisualStateWideMinWidth = -1 />
```

dg2k:
Indeed, it does! Didn't think of that at all and thanks.

The setting interface would still add some value to the Minimal template (of course, now taking account of  (HamburgerMenu.VisualStateWideMinWidth = value) and the modified PR follows.

PR #174, opened 2015-10-01
Changed Resizer into a ContentControl 

janabimustafa:


JerryNixon:
Impressively fast. Some cleanup is needed. Things like Register("Property") should be Register(nameof(Property)) so there are no literal strings. I also see changes to MainPage when that should not be changed. There are some default comments on your dp that should be cleaned up, too. Tweak it a little. That okay?

janabimustafa:
Yeah that's fine 😊 will work on it a bit more this weekend. Also, I will try to create a new ResourceDictionary for each control and reference it from generic.xaml

janabimustafa:
Also, I need to update my propdp snippet to stop using strings.

JerryNixon:
This PR is not compat with the master base, could you resubmit?

PR #176, opened 2015-10-01
Session state  change

waynebaby:
Found Session state using linear search and the api is not simple and not bindable.  

1  modified the property to 2 different property that can be access with both [] indexer and dynamic bag.
2  modified then implementation to IDictionary<string,object> based  ExpandoObject (supports INPC).

Hope this could be accepted.


waynebaby:
I noticed I submit too much  commits by dashboard.
the only useful commit is  this  0613a1a 
do I need recreate a new pull request?

JerryNixon:
Okay, work this out for me. StateItem<Type, Key, Value>. What about the scenario where the Value is null. How do you get the Type from null System.Object for to clear the value? Moreover, how could you prevent duplicating the dictionary entry when you write a value? Without Type, you would not be able to find the original null, right?

You need to be able to get the item by its Type, right? Even when null. I look forward to your reply because I might have misunderstood your intent here.

waynebaby:
- if value is null , the type means nothing. basicly it is not there.
- ExpandoObject is an Dictionay so it can be overwrited if the key are equal, so there will be no dupe
- get value by type just call `SessionState.Values.OfType<TType>()`. if you intent have multiple values in one key, and can give you a `Get<T>(key)` version of collection implementation still using Dictionary but that scenario  really smells bad.

waynebaby:
So how about I implement a version of collection have same method just like the old one, cored by dictionary ?

PR #177, opened 2015-10-01
AwaitableDelegateCommand added. can execute with a key and can be awaited by that key.

waynebaby:
A awaitable command  that can awaited by code by execution key string

### DIFF
--- a/Template10 (Library)/Mvvm/AwaitableDelegateCommand.cs
+++ b/Template10 (Library)/Mvvm/AwaitableDelegateCommand.cs
@@ -52,25 +52,55 @@ namespace Template10.Mvvm
 
         public void Execute(object p = null)
         {
+
+            string appKey = null;
+            InternalExecute(p, out appKey);
+        }
+
+        private bool InternalExecute(object p, out string appKey)
+        {
+            appKey = null;
             if (!CanExecute(p))
-                return;
+                return false;
             try
             {
                 var ap = p as AwaitableDelegateCommandParameter;
+                appKey = ap?.ExecutionKey ?? Guid.NewGuid().ToString();
+
                 Func<Task> e2 =
                    async () =>
-                     {
-                         await _execute(p as AwaitableDelegateCommandParameter);
-                         Task taskSelf;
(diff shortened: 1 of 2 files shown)


JerryNixon:
Show me the usage syntax.

waynebaby:
``` csharp
    public class VMMock
    {
        public VMMock()
        {


            MyAwaitableCommand = new AwaitableDelegateCommand(
                async cmdParameter =>
                {
                    await Task.Yield(); //<-- some async work 
                },
                cmdParameter => !IsBusy);

        }
        public bool IsBusy { get; set; }

        public AwaitableDelegateCommand MyAwaitableCommand { get; set; }


        public async Task Run3TimesInARow()
        {
            var commandParams = new[] { "exe1", "exe2", "exe3" }
            .Select(x=>new AwaitableDelegateCommandParameter { InnerCommandExecutionParameter ="whatever", ExecutionKey=x})
            .ToArray();

            foreach (var param in commandParams)
            {
                MyAwaitableCommand.Execute(param);
                await MyAwaitableCommand.AwaitByKey(param.ExecutionKey);
            }   //<--  can execute multiple commands in a row  and  use async/await avoid block ui thread.

        }
    }
```

waynebaby:
And the AwaitableDelegateCommandParameter can be contained in  a  resource dictionary to get datacontext and  bind  "ExecutinKey"  and "InnerCommandExecutionParameter"  to datacontext

waynebaby:
The problem in simple ICommand  and delegate implementation is  "Execute" leave no awaitables in return, the delegate is  void (object)  means  you can only waited in ui thread and this will kill your UX if you want to run long-life executing  combined command (Prism ->CombinedCommand).

if we use command like this, can solve the problem.  If you have mind in syntax please improve it.

JerryNixon:
Seems like this syntax would be easier, no?

``` csharp
Task ExecuteAsync(T obj);
```

waynebaby:
sure it could be `Task ExecuteAsync(T obj);`  I will add that. 

but await by key is still needed if the ui fired the command and await is called in VM.

JerryNixon:
I don't know.

PR #179, opened 2015-10-02
Implements a setting in Minimal template for  menu mode preference in…

dg2k:
Implements setting in Minimal template for menu mode preference in wide screen as CompactOverlay or CompactInline; normally only the latter is available. This is salvaged from PR #171 and this setting option adds to a list of user preferences.

### DIFF
--- a/Templates (Project)/Minimal/Services/SettingsServices/SettingsService.Apply.cs
+++ b/Templates (Project)/Minimal/Services/SettingsServices/SettingsService.Apply.cs
@@ -43,5 +43,12 @@ namespace Sample.Services.SettingsServices
         {
             Template10.Common.BootStrapper.Current.CacheMaxDuration = value;
         }
+
+        private void ApplyCompactMenuForWideScreen(bool value)
+        {
+            Views.Shell.ApplyCompactOverlayForWideVisual(value);
+        }
+        
+
     }
 }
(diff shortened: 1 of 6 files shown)


JerryNixon:
Why would the user do this @ runtime instead of the developer coding for it? I appreciate what you are doing here. And your PR clearly demonstrates a perfect understanding how Settings are handled in T10, but why? Are you sure this is worth adding to Minimal? Minimal?

dg2k:
I absolutely get your point. This contribution is within the developer domain (if I can call that) outside of where real contributions are expected. I can say I was a bit carried away -- as you'll recall, with the HamburgerMenu.VisualStateWideMinWidth = -1 magic unbeknown to me I did a bit of visualstate and dp coding,and liked the demo with the preference setting. I later tested with the HamburgerMenu.VisualStateWideMinWidth = -1  implementation and liked it even more and thought perhaps others may like it too (yah, getting carried away ;). It may be a better candidate as part of a sample? Whatever you think is appropriate ... I'm fine with it.

JerryNixon:
I tell you what. Let's not lose your code. Let's create a HamburgerMenu project in Samples and include this, but instead of in Settings, you can put it in MainPage. Make sense?

dg2k:
Thanks Jerry, shall do that. But know this: this is part of the code for an 'awesome app' I am developing, and as such nothing would be lost. I just wanted to share and in fact I have many like these that can feed into a sample. The setting in question  plus a 'status user notification' (which I haven't shared yet) make an excellent package for a sample. HamburgerMenu project sample it is. Consider this done.

PR #181, opened 2015-10-03
SessionState Collection Plan B+

waynebaby:
Hi  Jerry

Here is another implementation that I enhanced as dictionary.

I added some comment for some method.
please review.


waynebaby:
Hi J

``` csharp
return left?.Equals(right) ?? false;
```

this is not needed cos StateItemKey is a struct like `KVP<K,V>`. It won't be null.

JerryNixon:
This PR is not compat with the master base, could you update?

PR #188, opened 2015-10-04
Task from result abuse

waynebaby:
Return a new null valued Task<Object> in a  async method might cause some perf issule, and the code is hard to maintain correctly if we add more async calling in the method.  

I am trying to use Task.Yield() instead of Task.FromResult to avoid these in this pull requst


nguyenquyhy:
Is the perf issue really true? I follow sample codes and use Task.FromResult in many places in my projects. Please advise me more on this. Thank you very much.

JerryNixon:
Wayne. Could you demonstrate it?

waynebaby:
by ilspy in .net 4.5:

``` csharp
// System.Threading.Tasks.Task
[__DynamicallyInvokable]
public static Task<TResult> FromResult<TResult>(TResult result)
{
    return new Task<TResult>(result);
}
```

``` csharp
public static YieldAwaitable Yield()
{
    return default(YieldAwaitable);
}
```

new object vs default struct(const), default struct win.  (no set il executed).

The async function body acturally dose not need a real object to generate a yield, it need a awaitable stand there and it can jump to next state.

If you put a executed task in that line and await it, will do the same trick.

PR #189, opened 2015-10-04
Better bindable

waynebaby:
- make bindable  supports custom equal comparer
- can compare x,y then store to z now
- gether the logic in both INPC/DP bindable class to a single method.

usage:

``` csharp


    public class HighScoreSampleModel : BindableBase
    {
        public HighScoreSampleModel(int lastYearHighScore)
        {
            _highScroreLastYear = lastYearHighScore;
        }

        int _highScroreLastYear;

        public int HighScroreLastYear
        {
            get
            {
                return _highScroreLastYear;
            }
            set
            {
                Set(ref _highScroreLastYear, value, (old, v) => old >= v); //add compare rules.
            }
        }
        int _highScoreThisYear;

        public int HighScoreThisYear
        {
            get
            {
                return _highScoreThisYear;
            }
            set
            {
                Set(_highScroreLastYear, value, out _highScoreThisYear, (old, v) => old > v); //compare 2 fields and save to other field if needed
            }
        }
    }
```


waynebaby:
and this pull is requested in Tacoma airport.
going to buy some cock tail and get drunk before onboard.

PR #195, opened 2015-10-06
PageHeader theme binding fixed and defaults changed; theme switching …

dg2k:
PageHeader theme binding fixed and defaults changed; theme switching enhanced in Minimal template.

You'll note that from the outset PageHeader background was defined as Brush rather than SolidColorBrush, which means you have additional options such as LinearGradientBrush, RadialGradientBrush and ImageBrush to play with for more appealing PageHeader backgrounds (haven't thought much about  DrawingBrush and VisualBrush brushes but nothing stops you from trying out by defining in Custom.xaml style file). With this PR, I have included a LinearGradientBrush for Light Theme just to demo a different background brush.

PageHeader theme switching requires each page to update itself by reading updated background/foreground values in Shell properties (thru binding in respective viewmodel. Is there an automagic way that would force all instances of PageHeader to update globally without each page updating itself? Theme-aware usercontrol? Will PropertyMetadata PropertyChangedCallBack do such a trick? Haven't given much thought here and stopped myself digressing.

Theme switching (Dark/Light) while for FrameWorkElements is a single line code --  by setting the RequestedTheme property (in this case via Shell instance); for HamburgerMenu or PageHeader, on the other hand, it requires getting user defined values from Custom.xaml and apply to individual items. If there is a more elegant way of parsing user themes (perhaps new to Windows 10 which I've overlooked?) please let me know. An ideal scenario is whereby Windows somehow cooks Custom.xaml resources that leads to an object providing easy access to Dark, Light, Default resources by theme category ... but haven't seen one. (I mean if you can do better than parsing with what XAML provides with ResourceDictionary.ThemeDictionaries).

### DIFF
--- a/Template10 (Library)/Controls/PageHeader.xaml
+++ b/Template10 (Library)/Controls/PageHeader.xaml
@@ -10,17 +10,12 @@
              x:Name="ThisPage" MinHeight="48"
              d:DesignWidth="400" mc:Ignorable="d">
 
-    <UserControl.Resources>
-        <SolidColorBrush x:Name="DefaultHeaderBackground" Color="Gainsboro" />
-        <SolidColorBrush x:Name="DefaultHeaderForeground" Color="Black" />
-    </UserControl.Resources>
-
     <CommandBar x:Name="HeaderCommandBar" Grid.Column="0"
                 Grid.ColumnSpan="2"
-                Background="{x:Bind HeaderBackground, FallbackValue=Blue}"
+                Background="{x:Bind HeaderBackground, FallbackValue=Gainsboro, Mode=OneWay}"
                 ClosedDisplayMode="Compact"
                 DataContext="{Binding ElementName=ThisPage}"
-                Foreground="{x:Bind HeaderForeground, FallbackValue=White}">
+                Foreground="{x:Bind HeaderForeground, FallbackValue=Black, Mode=OneWay}">
 
         <VisualStateManager.VisualStateGroups>
             <VisualStateGroup x:Name="VisualStateGroup">
--- a/Template10 (Library)/Controls/PageHeader.xaml.cs
+++ b/Template10 (Library)/Controls/PageHeader.xaml.cs
@@ -72,20 +72,20 @@ namespace Template10.Controls
 
         public Brush HeaderBackground
         {
-            get { return (Brush)GetValue(HeaderBackgroundProperty) ?? Resources["DefaultHeaderBackground"] as Brush; }
+            get { return (Brush)GetValue(HeaderBackgroundProperty) as Brush; }
             set { SetValue(HeaderBackgroundProperty, value); }
         }
         public static readonly DependencyProperty HeaderBackgroundProperty =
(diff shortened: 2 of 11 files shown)


janabimustafa:
I can see this becoming a nightmare to maintain. This could be solved by converting the PageHeader control to a Templated ContentControl instead of UserControl, and that's what I am trying to do right now. I am at a roadblock currently because of the fact that PrimaryButtons and SecondaryButtons properties are read-only on the CommandBar control (so I can't really setup binding with them). Any help @JerryNixon ?

dotnetnoobie:
You can create DependencyProperty's for the two collections of buttons
then you can bind to them in your Templated ContentControl 

then when they change you can set the them to the 
CommandBar.PrimaryButtons or SecondrayButtons via the "PropertyChangedCallback" of the DependencyProperty, or if you know they will always be set when the control is created you can do it in the Constructor 

That is if I understand your question correctly?

You need to loop over the DependencyProperty's for each button collection and use the 
PrimaryCommands.Add(<your button>); method

janabimustafa:
We can't set them in the constructor (Like the way it is done now) as the template only applies later. As for creating these two dependency properties, I am having a problem making the control accept a collection of elements. 
For example

``` xaml
<mycustomcontrol>
  <mycustomcontrol.PrimaryButtons>
    <somechildcontrol/>
    <somechildcontrol/>
  </mycustomcontrol.PrimaryButtons>
</mycustomcontrol>
The DependencyProperty of the PrimaryButtons has type IObservable<somechildcontrol> or List<somechildcontrol> or even a specific ObservableCollection<somechildcontrol>
```

This will give me an exception of "cannot set somechildcontrol to collection of somechildcontrol."
Do you know what's the proper way to handle that?

dotnetnoobie:
I made my own control which is inherited from the SplitView and has a Hamburger menu and the Commandbar/Page header all in 1.

So I have two Controls, a ShellMenu control and a ShellMenuItem Control 

in the ShellMenuItem Control I used Dependency Property of IList<> to add the CommandBar items I want to make the CommandBar Buttons from.

I tried to use the "ICommandBarElement" but when I did this I could not get the "Command" binding to be passed onto my Control from xaml...

so I made an custom object and put the command name (as a string) as a property 
then in my control I did the binding in C# using the command name "string" (this is new really a very good solution)

in the constructor of the ShellMenuItem I set the IList<ShellCommandBarItem> dependency properties for Primary and Secondary buttons to new List<ShellCommandBarItem>

Then I am able to add the buttons (really ShellCommandBarItem objects)  in xaml,
then in the ShellMenu control I create the AppBarButtons from the properties of these objects and add them to the Primary and Secondary button collections of the CommandBar in the ShellMenu Control using the "Add" method

janabimustafa:
Thank you for that example. I was forgetting to set the PrimaryCommands to an instance of a collection in the constructor so instead of adding these elements in the xaml, it was trying to set the PrimaryCommands to the first one :)

dg2k:
Closing this PR. See my remarks in #202.

Issue #175, opened 2015-10-01
PageHeader Color Theme


dg2k:
There is something different about PageHeader - you cannot change its background or foreground colors directly, and even the Darlk/Light theme setting has no effect on it. On the contrary the HamburgerMenu color changes behave as you'd expect.  Is this by design? Didn't think so. Just thought I should ask if there is a 'by-design' issue before I look into it.

Couple of weeks ago I tried to change the color theme programmatically, which is now done for us in the Minimal template, and that was when noticed PageHeader didn't change. The same issue exists with the theme setting already implemented and the question is still relevant.

JerryNixon:
> On the contrary the HamburgerMenu color changes behave as you'd expect. Is this by design? 

Yes. :) The HamburgerMenu has gotten significantly more love. This is a good issue report. I am going to ask you to suggest the fix. I am about to step out on vacation for a week. The new AccentColor on the HamburgerButton is not relevant on the PageHeader (you didn't say it was) but implementing that caused the HamburgerMenu to get tested over and over until it worked as expected. PageHeader needs the same attention.

dg2k:
Re-reading my posting, I note some ambiguity. The question "Is this by design? Didn't think so" was intended to refer to the PageHeader, not the HamburgerMenu. At any rate, no info was lost as we understand what needed fixing. Yes, HamburgerMenu behaves as expected, even when I was messing with it to change the theme dynamically.
Have a nice break Jerry!

JerryNixon:
Can't wait to see your PR.

dg2k:
See PR #195.
Update: See my comment in PR #202.

Issue #200, opened 2015-10-09
Hamburger Menu Colors - Unable to see selected item


darenm:
![hamburgercolorswrong](https://cloud.githubusercontent.com/assets/1674843/10406095/6643b3f6-6e92-11e5-9c54-fb959c10c2ab.gif)

dg2k:
The HamburgerMenu and PageHeader have a few issues I have noticed. PR #195 will, for the most part, address these issues (but lately noticed further outstanding style corrections to be made). There are a number of things confusing the issue here: 1) when Custom.xaml was split and then recombined, I think, somethings got changed/lost on the way (sorry if this sounds a blame, it isn't); 2) the AccentColor which I believe is meant for testing?? does some overriding/changing of color schemes so Custom.xaml definition may not appear as it should.

I know PR #195 will take us a step closer to addressing the style issues and then Custom.xaml should work as expected.

I checked if modifying Custom.xaml alone fixes the problem but didn't work without the fixes in PR #195 (also noting that AccentColor has some impact).

dg2k:
Update: Please see the remark in #202. We better address styling issues after that [major] merge.

PR #202, opened 2015-10-11
Changed page header into a templated control

janabimustafa:
This pull request introduces the new PageHeader control that is customizable. It also fixes the theming issue with the old control. Next up is the HamburgerMenu control.
Addressing issue #175


dg2k:
This is a great contribution and I shall close down PR #195 in favour of this.

I tended to steer away from CustomControl but I needn't as this really demonstrates the power templated control. The PageHeader template control brings order as there is no updating required by individual pages (which btw I have checked). Whatever remains of my contribution, I shall do a PR after this one; for instance, there are some styling issues (e.g. Daren's remark in #200, be it more of a developer issue) and shall include such like after this.

JerryNixon:
Can you sync HamburgerMenu to take the same style approach? Don't mess with changing it to a templated control just yet. Make sense? Thanks.

PR #207, opened 2015-10-12
Applying same fix to ForwardRequested as we did in BackRequested to h…

bc3tech:
…andle users subscribing to ForwardRequeted but not setting Handled = true


bc3tech:
See #206

PR #230, opened 2015-10-14
Inherit PageHeader from CommandBar. …

MichaelPetrinolis:
Better Design Time support. 
Removed Text dependency property and reintroduced the content property. 
Ellipsis Button is always visible, following the commandbar behavior.

We can hide the ellipsis if no commands exist, or introduce a behavior property to control it


JerryNixon:
Good idea. Could you re-merge and resubmit? This is out of sync.

PR #231, opened 2015-10-15
Pageheader inherits from CommandBar

MichaelPetrinolis:
Better Design Time support. 
Removed Text dependency property and use commandbar's content property (template and template selector) to provide the Header's Content
The ellipsis now is always visible following the commandbar behavior.
DisplayMode compact behaves like the previous version of PageHeader


JerryNixon:
Do I like this new approach? I love this new approach. But your PR is out of sync with the code base. Could you resubmit it?

JerryNixon:
DANG! I just realized I did this to you two times. I just committed the 1.0.5 Nuget changes and they were too many before I merged. Dude, I am sorry. Please resubmit and don't hate me.

Issue #163, opened 2015-09-29
SplitPaneView in HamburgerMenu doesn't render properly when Page.BottomBar has app/command


bc3tech:
The bottom part of the hamburger is cut off by the height of the command/app bar in the app.

JerryNixon:
Is the CommandBar in Shell or MainPage?

bc3tech:
@JerryNixon MainPage

JerryNixon:
You may have discovered a bug in the SplitView. 
I will need to investigate in a week after I return.

dg2k:
Please see PR #233.

JerryNixon:
I created issue #246 to correct this. You are welcome to try to implement.

PR #233, opened 2015-10-15
Addresses issue in #163 regardng Page.BottomAppBar. May not be a perfe…

dg2k:
First of all, Page.BottomAppBar covering the page sounds like a bug in XAML?? as a quick search online suggests this. I cannot for sure say this is the case.  Having read issue #163, I tried a workaround on my system, and it worked with the previous version of ApplyAppTheme() in SettingsService.Apply.cs (in Minimal template), but  with the current version of ApplyAppTheme() that calls RefreshStyles() some anomaly will be introduced (arising from the very solution attempted). I am making this PR just in case this leads to a perfect solution. At any rate, anyone using Page.BottomAppBar is welcome to make use of this code as it is fully working with an older version of ApplyAppTheme() in SettingsService.Apply.cs. I am yet to figure out how ApplyAppTheme() with RefreshStyles() is different and perhaps Jerry could help on this one.

The suggested solution is as follows: A DP is defined to set a boolean when a page has BottomAppBar (set true in OnNavigatedTo and false on NavigatingFrom). The DP sets a visual state that moves the SplitView up by 20px (to avoid covering by BottomAppBar) or restores to 0 if false.

There may be a smarter and more concise solution but this is what I could come up with.
I thought of if the SplitView's Pane and Content could be positioned at ease (this could solve the anomaly referred to and hence problem solved) but from the code can't see any easy way for this...

PS: A page with BottomAppBar is added to demo the suggested solution.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -168,6 +168,10 @@
     </UserControl.Resources>
 
     <Grid x:Name="RootGrid">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="*" />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
 
         <VisualStateManager.VisualStateGroups>
             <VisualStateGroup x:Name="VisualStateGroup">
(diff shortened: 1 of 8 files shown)


JerryNixon:
Yikes, this is a brutal approach. I like what you did in shell, just not the extra page. Let's come up with a better solution. I am 100% ready to make the developer somehow handle this, too. I think this might be better as a Sample, too. Maybe. This is a crappy bug, huh?

Issue #228, opened 2015-10-14
Unit testing view models ?


dpakpaul:
How do I write unit tests against my view models since ViewModelBase has the NavigationService, Dispatcher and SessionState properties that I cannot mock or inject ?

JerryNixon:
Well, until we setup the three services as interfaces and update INavigable to use them, there's no way to properly test your view-models. I look forward to your pull request. :)

PR #236, opened 2015-10-15
Issue #228 - make BaseViewModel instances DI compliant

dpakpaul:
This is a PR for issue #228 . The fundamental issue was that the BaseViewModel had 3 concrete instances that it had a dependency on - the NavigationService, the DispatcherWrapper and the StateItems dictionary.

Extracted interfaces for all 3 cases. 
Added these to INavigable
Implementations are injected in the OnNavigated() in NavigationService.

Side note - while looking at the unit tests project (temporarily removed from sln?), I see that the unit tests for the StateItems class are referring to an older definition of the class. Also, I see issues in the StateItems class implementation as well - particularly in the Remove(object value) method - will be a endless recursive call. Will put in a separate issue on that ?

### DIFF
--- a/Samples/MvvmLight/Mvvm/ViewModelBase.cs
+++ b/Samples/MvvmLight/Mvvm/ViewModelBase.cs
@@ -13,12 +13,12 @@ namespace Sample.Mvvm
         public virtual void OnNavigatingFrom(Template10.Services.NavigationService.NavigatingEventArgs args) { /* nothing by default */ }
 
         [JsonIgnore]
-        public Template10.Services.NavigationService.NavigationService NavigationService { get; set; }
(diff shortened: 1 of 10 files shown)


JerryNixon:
Will need to think it over if the interfaces should be moved to an interfaces namespace. Do you have any thoughts on that?

dpakpaul:
I think interfaces and implementations should be in functional groups under the same namespace rather than in separate namespaces. So, for example,  I would rename the namespace "Template10.Services.NavigationService" to just "Template10.Services.Navigation" since the additional "service" is redundant and clarifies the functional categorization of the namespace. Then the interfaces and implementations and other code entities related to navigation live in that same namespace. Also reduces the "usings" in source files.

As an aside - the Prism project recently did exactly this and got rid of their interfaces namespaces.

PR #245, opened 2015-10-16
Implementing a more deliberate handling of stateful navigation parameters…

bc3tech:
The previous error message to Override ToString() isn't valid; ToString() always exists, so clearly that's not the problem when an exception is thrown.

The real problem is the Settings containers only accept [a certain subset of types](https://msdn.microsoft.com/en-us/library/windows/apps/xaml/mt299098.aspx#types_of_app_data) (and do _not_ automatically `ToString()` stuff that doesn't fall in to those).

Instead, I opted for the code to `try` and throw the nav param in to the collection; if that fails, check to see if the consumer has implemented a new interface denoting our parameter is stateful. In addition to providing a way to get the state from an object like this, it requires the developer implement a way to restore from the saved state as well along with direction (via XML comments) on how to make that happen.


JerryNixon:
I thought the prev implementation would attempt to serialize everything. We want the developer to save anything and only get an error if we can't serialize or Parse it. This make sense?

JerryNixon:
We definitely do not want the developer to have to do anything special unless they are saving to settings something that is simply unable to be Parsed or serialized.

JerryNixon:
Looking more closely at the changes you have committed, I can tell, I don't understand what you are solving here and what's going on. Be patient with me.

bc3tech:
The issue is the usage of any object - as a nav param - that `LocalSettings` auto-serialization doesn't support, since the "bag" used to hold state information is based on `LocalSettings`.

First and foremost, the exception thrown in this case says the dev should override `ToString()` to fix it; this isn't correct. Doing so doesn't help the situation (and why would it? ToString() is _always_ available, right?).

Instead, my thought is that if a user wants to save such data, they should be able to if they provide the necessary facilities, which is what this attempts to enforce. If a user just passes strings around, this change is invisible to them. If they pass their own objects, then they've gotta use this interface or they'll get an exception at serialization time (which is more accurate than the exception thrown today telling them to override `ToString()` which in fact does _not_ fix the issue)

JerryNixon:
What about an override:

```
T Settings.Read<T>(key, Func<string, T> parse, T otherwise);
Settings.Write<T>(key, T value, Func<T, string> unparse);
```

This would give you what you want?

bc3tech:
it would, but wouldn't work (as automatically) during our assign of the State variables during suspension.

even if you (we) don't want to implement my solution, at a minimum the exception text needs to be changed.

JerryNixon:
Do you still feel this is relevant now that settings uses a converter?

bc3tech:
haven't pulled latest in a while and we're working with a version of the library that uses this solution now.

I'll have a look one of these days and let you know; no promises on timeline though.

PR #248, opened 2015-10-17
Set method for properties, clean

ivesely:
Set method for properties
Namespace Repositories dont exist and Sample.DataModels clean(twice using Sample.Shared)


JerryNixon:
Oops, you added the methods two times. Also, be sure your argument variables' first character is lower case. I am not going to accept this one, please re-submit with the fixes.

ivesely:
Hi Jerry, it's two times because one time for BindableBase and one time for DependencyBindableBase. I will fix the other problems.

Issue #246, opened 2015-10-16
Update HamMenu to react to Frame.Page.BottomAppBar close/open


JerryNixon:
The bug right now is that if you use Page.BottomAppBar (or TopAppBar) instead of a page-level grid, then when it is rendered it hides the Pane portion of the HamMenu which, if it has secondary buttons, will make them either unusable or invisible.

> It is reasonable for some apps to use the BottomAppBar because it slides up with the SIP. I believe this is a very distant edge case, but still worth correcting for. 

The solution is to have the HamMenu listen for Page navs and when there is a new page, it should ask the page for its BottomAppBar and if it is set then handle either AppBar or CommandBar. In that handler we will adjust the top or bottom margin of the Ham.Pane. 

Note: it will be very important that the unwire occur so we don't have a memory leak. A better and possibly easier solution would be add WeakReference<T> to T10 library (pref in the Mvvm folder) and use that to attach so we don't have to un-attach later.

dg2k:
See PR #252.

JijoBose:
I have a workaround for this issue here.

https://github.com/Windows-XAML/Template10/issues/406#issuecomment-228585169

PR #252, opened 2015-10-18
A solution to Issue #246: Update HamMenu to react to Frame.Page.Botto…

dg2k:
Hope this does the trick.
Easier than I thought by checking for Page.BottomApBar page property.


dg2k:
Resubmitted under PR #260. Please read the detailed comment.

PR #253, opened 2015-10-18
PageHeader title left margin set to 16px as per original setting for …

dg2k:
PageHeader title left margin is set to 16px as per original setting for PageHeader. Otherwise you'll note that the PageHeader title text touches the HamburgerMneu pane without a gap.


SimoneBWS:
You are right, but at this point I think that is better if we add ContentMarginProperty

MichaelPetrinolis:
We could also bind the margin of the contentpresenter to the padding property of the pageheader. You could submit a PR

PR #260, opened 2015-10-19
A solution to Issue #246: Update HamMenu to react to Frame.Page.Botto…

dg2k:
<h3>How the solution works:</h3>

**Detection:** When a page is navigated to (detected in the navigation base class NavigationService.cs) the property page.BottomAppBar is checked for. If it's non-null && Visible, it means we have a BottomAppBar. The page.BottomAppBar?.Visibility condition sets a boolean DP PageHasBottomAppBar in HamburgerMenu accordingly.

**Operation:** The DP PageHasBottomAppBar listens for the detection boolean value as above and accordingly chooses a visual state for page.BottomAppBar existence or absence (the default state).
When page.BottomAppBar is present, the visual state makes a 20px high rectangular spacer 'Visible' and this pushes the HamburgerMenu pane up by 20px, else the rectangular spacer collapses (to default visual state).

The developer needs to set BootStrapper.Current.HamburgerMenu to HamburgerMenu object in Shell constructor. This is already required for ApplyAppTheme() anyway and as such not an extra requirement for Minimal template except that the HamburgerMenu property is moved from Shell to BootStrapper.

**What happens if** the developer doesn't (or forgets to) set BootStrapper.Current.HamburgerMenu  property to HamburgerMenu object?
Nothing. It means if page.BottomAppBar exists, the partial covering of HamburgerMenu pane would be noticed. Transparent to the developer, it is.

I will be posting a Sample that demos page.BottomAppBar and other features of HamburgerMenu distilled from my recent PR attempts.

### DIFF
--- a/Template10 (Library)/Common/BootStrapper.cs
+++ b/Template10 (Library)/Common/BootStrapper.cs
@@ -423,5 +423,7 @@ namespace Template10.Common
                 return _PageKeys as Dictionary<T, Type>;
             return (_PageKeys = new Dictionary<T, Type>()) as Dictionary<T, Type>;
         }
+        public Controls.HamburgerMenu HamburgerMenu { get; set; }
+
     }
 }
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -168,6 +168,10 @@
     </UserControl.Resources>
 
     <Grid x:Name="RootGrid">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="*" />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
 
         <VisualStateManager.VisualStateGroups>
             <VisualStateGroup x:Name="VisualStateGroup">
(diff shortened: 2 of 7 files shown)


JerryNixon:
You said

> The developer needs to set BootStrapper.Current.HamburgerMenu to HamburgerMenu object in Shell constructor. This is already required for ApplyAppTheme() anyway and as such not an extra requirement for Minimal template except that the HamburgerMenu property is moved from Shell to BootStrapper.

This is already required? Where?

dg2k:
Perhaps 'required' is not a correct word for my remark, but the following will clarify.

Note the following in Shell.xaml.cs:

`public static HamburgerMenu HamburgerMenu { get { return Instance.MyHamburgerMenu; } }`

I am referring to ApplyAppTheme() in SettingsService.Apply.cs for Minimal template (which in turn is used to call the RefreshStyles() in HamburgerMenu  by using the above).

My remark is simply intended to emphasize the fact that an existing property was used and not created a new one to get to a HamburgerMenu object. The one modification I made though is that I moved the HamburgerMenu property you defined in Shell to BootStrapper so that it would be easily accessible to both SettingsService (in Minimal) and NavigationService (in Library, where BottomAppBar is detected). Hence, the above property in Shell is now replaced with:
 `BootStrapper.Current.HamburgerMenu = Instance.MyHamburgerMenu;`
in the Shell ctor().

Regardless of the above explanation, the solution puts no burden on the developer; if the developer notices the BottomAppBar problem, then, they must ensure:

 `BootStrapper.Current.HamburgerMenu = Instance.MyHamburgerMenu;`

in the Shell ctor(). It's a solution that guides the developer take a specific action, else everything is the default situation.
An alternative approach _may be_ possible: solve the problem without the developer even noticing the BottomAppBar problem. In this case, we can initialize `BootStrapper.Current.HamburgerMenu` behind the scene rather than Shell ctor() do it for us. Would this be preferable? I can see one advantage in this: if the platform bug with BottomAppBar is removed, then the 'solution code' can be removed without the developer's knowledge. But, we still need the Shell instance ...

JerryNixon:
I think the solution should be in the HamburgerMenu itself, querying its Frame property and attaching to Nav events. This would allow you to use the HamMenu without knowing that it was listening. Without creating any properties. Without doing anything. The HamMenu would simply be listening for the Page to change and when it does it would ask for Top and Bottom bars, attaching to those as needed and responding to their visibilities. This seems to me to be the most developer low-tech approach. Which is my goal. Does this make any sense though?

dg2k:
I will give it some thought.

dg2k:
Done!
See PR #273.

Issue #262, opened 2015-10-20
Splash screen modal


youtoofan:
I'm not sure if this is intentional, but while the splashscreen is loading the shell is loaded too.
You can test this by setting the Task.Delay value to >2500 ms in the Minimal sample.

After the initial splash logo, the splash screen shows up, immediatly followed by the Shell while the splash is still loading.

I want the splash screen to remain modal, because some background data is being transfered at that point from the azure service.

JerryNixon:
I do too. Any suggestions?

youtoofan:
Funny thing is, I don't think I had this behaviour with version 1.0.3 of the library.
I'm thinking of using the Shell.IsFullScreen property for this. #263 will have to be fixed first though.

JerryNixon:
Fixed. You mean pushed to NuGet?

callummoffat:
I developing a fix for this.

callummoffat:
It involves adding a view model to Splash, which handles enabling the IsFullScreen property for Shell's HamburgerMenu instance via OnNavigatingTo & OnNavigatingFrom in the ViewModel. This is done by adding a SetFullScreen(bool) static method in Shell and activating it in the methods.

callummoffat:
My implementation involves adding a static SetFullScreen(bool) to Shell, which toggles the property in MyHamburgerMenu. I then call this method as soon as the Shell object is initialized in App to enable full screen mode, and again when the main page is navigated to disable.

JerryNixon:
If you remove IsFUllScreen=False from XAML does that do the same thing?

callummoffat:
I'll look into that. Just testing now.

callummoffat:
I can't tell. Splash screen flashes by too fast for me to tell.

callummoffat:
I'll try adding a breakpoint, and test again.

callummoffat:
I still can't tell. Splash doesn't seem to be displaying. Either that, or it's popping through the implementation too fast. I'll try with my app which uses Template 10.

callummoffat:
Yeah, that doesn't work. Tested that in my app, and it stops the menu from displaying. However, it keeps the menu full screen after the splash has completed. So, it's probably necessary to use the method I developed.

JerryNixon:
This is a different error. Full screen should be automatically toggled of if 1. The navigation service navigated and 2. Full screen has not been manually set. 

Hmm. Will need to review.

PR #272, opened 2015-10-23
Initial implementation of fix for issue #262

callummoffat:
This is a fix for issue #262. It works by adding a method in Shell to toggle the HamburgerMenu's IsFullScreen property. This is then called in Minimal's App class after the Shell is init'ed, and again after MainPage has been navigated to.


JerryNixon:
Feel like this (intended behavior) is a better solution:

> This is a different error. Full screen should be automatically toggled of if 1. The navigation service navigated and 2. Full screen has not been manually set.

callummoffat:
This is not needed.

PR #273, opened 2015-10-23
A solution to Issue #246 (Take 2): Update HamMenu to react to Frame.P…

dg2k:
This is a revised solution of PR #260 with code now confined to HamburgerMenu (see background info on this in PR #260).

More insight into `HamburgerMenu` code reveals that what was done previously with `NavigationService` (in NavigationService.cs) and DP in `HamburgerMenu` could actually be done by subscribing to the `NavigationService.FrameFacade.Navigated` event within `HamburgerMenu`; this only required tapping into `HighlightCorrectButton()` event handler which is already a subscriber.

Thank you Jerry for the remark in PR #260 - because of that now I am getting to know the innards of `HamburgerMenu`. The previous spaghetti code is now a neat minimalist single if-block in `HamburgerMenu` codebehind.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -168,6 +168,10 @@
     </UserControl.Resources>
 
     <Grid x:Name="RootGrid">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="*" />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
 
         <VisualStateManager.VisualStateGroups>
             <VisualStateGroup x:Name="VisualStateGroup">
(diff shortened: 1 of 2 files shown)


JerryNixon:
Innards. Sounds gross. :)

dg2k:
Gross only if taken out of context (i.e., alternative definition referring to an animal? ;-)
Merriam-Webster definition: _the internal parts especially of a structure or mechanism._

Any comment on the proposed solution?

JerryNixon:
Tell me how it will react when the user clicks the CommandBar to change the size?

dg2k:
A 20px rectangular spacer un-collapses to push the HamburgerMenu pane so as to avoid covering by CommandBar. Happy to post a sample project to demo this in action.

JerryNixon:
We need a sample to demonstrate this for sure. You can handle that?

JerryNixon:
Also, can we change the VS to use setters instead of storyboards?

JerryNixon:
No. This is just not the right approach.

dg2k:
Okay. I would really love to know a better solution to this.

JerryNixon:
https://github.com/Windows-XAML/Template10/commit/e465b55a596138f77c75de4b223a4b24e80f3bdf

Issue #258, opened 2015-10-18
Comparing with LastNavigationParameter


pekspro:
In NavigationService there is this method:

```
    public bool Navigate(Type page, object parameter = null, NavigationTransitionInfo infoOverride = null)
    {
        if (page == null)
            throw new ArgumentNullException(nameof(page));
        if (page.FullName.Equals(LastNavigationType)
            && parameter == LastNavigationParameter)
            return false;
        return FrameFacade.Navigate(page, parameter, infoOverride);
    }
```

I suggest to change this to:

```
    public bool Navigate(Type page, object parameter = null, NavigationTransitionInfo infoOverride = null)
    {
        if (page == null)
            throw new ArgumentNullException(nameof(page));
        if (page.FullName.Equals(LastNavigationType))
        {
            if (parameter == LastNavigationParameter)
                return false;

            if (parameter != null && parameter.Equals(LastNavigationParameter))
                return false;
        }

        return FrameFacade.Navigate(page, parameter, infoOverride);
    }
```

I had a problem with this in my application where I generated a new string every time I wanted to go to a page. Then the check (parameter == LastNavigationParameter) failed since it just checked the objects.

Normally I would do a pull request, but this time I’m lazy :-).

JerryNixon:
Also parameter should check with isequals too right?

pekspro:
Yes, didn’t I do that? Or am I missing something? :-)

JerryNixon:
Isn't this your code?

```
if (parameter == LastNavigationParameter)
```

pekspro:
Yes and no :-). It’s from the original code I just moved it.

That check is still needed in cases where both parameter and LastNavigationParameter is null or if they actually are equal. The other check should cover all other cases

PR #294, opened 2015-11-01
Fix issue #258 and bugs with tiles

pekspro:
This fixes issue #258 and various minor bugs and improvements related to tiles.


JerryNixon:
I will merge this but the technique in the tile sample is the preferred tech. Honestly we should probe remove this from services someday. Not now though. Don't panic. :)

PR #296, opened 2015-11-03
FIX Settings can't be read.

lukaszkierepka:
Settings can't be read since new implementation of Write and Read in PropertyStore.cs.
In the read function it's trying to deserialize Values[key].ToString() but it's an ApplicationDataCompositeValue.


JerryNixon:
Excellent work.

PR #297, opened 2015-11-03
added BackOtherwise property to ValueWhenConverter in order to support ConvertBack

MichaelPetrinolis:
Also added an example in Converters Sample.


MichaelPetrinolis:
After a second thought, a better name would be OtherwiseBackValue instead of BackOtherwise.
If it is preferred, I could change the PR.

seesharprun:
It's difficult to come up with a simple name that conveys what you are doing even though it logically makes sense.

MichaelPetrinolis:
I agree, @seesharprun. I will close this PR and make a new one, renaming the Property to OtherwiseValueBack  (in order to follow the ConvertBack naming in IValueConverter interface), raise an exception when converting back and no value is set to OtherwiseValueBack and also check passed parameter in the Convert function

PR #301, opened 2015-11-04
Adding Set and RaisePropertyChanged variations to BindableBase

Delaire:
Added variations of Set and RaisePropertyChanged in the BindableBase
class so that properties written as follow can work:

``` csharp
public bool _isVisible;
public bool isVisible {
    get {
      return _isVisible;
    }
    set {
      Set(() => isVisible, ref _isVisible, value);
      RaisePropertyChanged(() => Title);
    }
}
```

This will allow developers that have used these properties to not have
to rewrite the code if they are migrating there application to
Template10 (like I am)

### DIFF
--- a/Template10 (Library)/Mvvm/BindableBase.cs
+++ b/Template10 (Library)/Mvvm/BindableBase.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
 using System.Linq.Expressions;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using Windows.UI.Xaml;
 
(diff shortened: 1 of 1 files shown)


Delaire:
tested and working now, fixed null exception

MichaelPetrinolis:
Hi @delaire, maybe this should be a migration sample (in MVVM for example) instead of complicating template's 10 base classes. Template 10 should be Spartan, as simple as it gets.

JerryNixon:
Okay, but use your favorite MVVM framework instead of T10.

Delaire:
@JerryNixon  small question, why should we use another Mvvm Framework (MvvmLight for example) instead of the one that is in T10?

@MichaelPetrinolis yes I have added this because of a migration that I am doing =)

PR #302, opened 2015-11-04
Removed redundant BackRequested handling in FrameFacade.

FrayxRulez:
Handle BackRequested if args.Handled == false breaks scenarios where you have two or more inner frames and you have to handle back in different ways depending on the situation.

### DIFF
--- a/Template10 (Library)/Services/NavigationService/FrameFacade.cs
+++ b/Template10 (Library)/Services/NavigationService/FrameFacade.cs
@@ -35,10 +35,10 @@ namespace Template10.Services.NavigationService
                 BackRequested.Invoke(this, args);
             }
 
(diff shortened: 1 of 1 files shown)


JerryNixon:
Then what will call Back() for developers who do not implement a custom handler?

FrayxRulez:
If you read at line 244 of BootStrapper you can see that root frame GoBack is already called. I think that secondary frames in most of situations requires a custom navigation handling.

https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Common/BootStrapper.cs#L244

Edit: Maybe a check on BootStrapper.BackButton.Ignore before commented code execution can solve the problem?

JerryNixon:
Want to resubmit?

FrayxRulez:
Done! https://github.com/Windows-XAML/Template10/pull/314

PR #318, opened 2015-11-10
PaneOpen and PaneClosed events: PanelClosed renamed PaneClosed for naming consistency.

dg2k:
PaneOpen and PaneClosed events: PanelClosed renamed PaneClosed for naming consistency.


JerryNixon:
Good one.

PR #320, opened 2015-11-10
Example in converters to show x:Bind to an ItemSource control 

MichaelPetrinolis:
with two way compiled binding on SelectedItem. Also changed ChangeTypeConverter not to throw must implement IConvertible when converting back to System.Object


JerryNixon:
Many compliments.

PR #323, opened 2015-11-11
Added PageTitleVisibility property to PageHeader.This is required to …

dg2k:
For PaneOpen and PaneClosed events to work as intended, a PageTitleVisibility property is required to work in conjunction with page Visual States.

**Event subscription example in a viewmodel ctor():**

`Views.Shell.HamburgerMenu.PaneOpen += (s, e) => { PageVisualState = "ShowTitleState"; };
             Views.Shell.HamburgerMenu.PaneClosed += (s, e) => { PageVisualState = "NarrowState"; };`

**Visual State setter fragment in XAML (ShowTitleState):**
`<Setter Target="pageHeader.PageTitleVisibility" Value="Visible"/>`

**Visual State setter fragment in XAML (NarrowState):**
`<Setter Target="pageHeader.PageTitleVisibility" Value="Collapsed"/>`

In View page, **DataTriggerBehavior** listens to a **PageVisualState** property in ViewModel to effect the **GoToStateAction** for visual state transition.

### DIFF
--- a/Template10 (Library)/Controls/PageHeader.cs
+++ b/Template10 (Library)/Controls/PageHeader.cs
@@ -37,6 +37,17 @@ namespace Template10.Controls
         public static readonly DependencyProperty BackButtonVisibilityProperty =
             DependencyProperty.Register(nameof(BackButtonVisibility), typeof(Visibility), typeof(PageHeader), new PropertyMetadata(default(Visibility)));
 
+        
+
+        public Visibility PageTitleVisibility
+        {
+            get { return (Visibility)GetValue(PageTitleVisibilityProperty); }
+            set { SetValue(PageTitleVisibilityProperty, value); }
+        }
+
+        public static readonly DependencyProperty PageTitleVisibilityProperty =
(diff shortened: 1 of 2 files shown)


JerryNixon:
No, it seems like you should do this instead:

``` csharp
<controls:PageHeader>
    <controls:PageHeader.Content>
        <TextBlock x:Name="MyPageTitle" Visibility="Collapsed">Jerry</TextBlock>
    </controls:PageHeader.Content>
```

Too many properties makes the control too complex. Any reason why not to do it like this (above)?

dg2k:
The net effect is the same, but with PageTitleVisibility you don't have to pull in a TextBlock content into PageHeader. In that sense,  I suppose, easier for the developer and neater. I agree that too many properties are not desirable but PageHeader is not that complicated (in my opinion).

For instance, going thru my implementation, all I have on PageHeader is localized strings via x:Uid settings and I don't use any TextBlock for page titles.

MichaelPetrinolis:
I aggree with @jerrynixon too. A dev could  retemplate the control and add a controlsize adaptive trigger to it, so if it's width become less than x hide the content presenter.also a behavior could do the same. Maybe this should not be an out of the box functionality for pageheader

JerryNixon:
Good discussion and it's not a bad idea. Let's minimize T10.

PR #335, opened 2015-11-13
Fixes for multiple nav services created after suspend/resume on mobile

### LINKED ISSUE #325: WP10: App stuck at Splash.xaml, if the app is open and the user launches the app again
I have no idea where the bug is, normally the app should jump on the MainPage.

tested with minimal project and WP10 Build 10581/10586.
splashscreen

bc3tech:
Fixes #325 by clearing all cached navigation services when the app is suspending since these are **re**-created upon resume

### DIFF
--- a/Template10 (Library)/Common/BootStrapper.cs
+++ b/Template10 (Library)/Common/BootStrapper.cs
@@ -299,7 +299,15 @@ namespace Template10.Common
         /// because the asunc operations are in a single, global deferral created when the suspension
         /// begins and completed automatically when the last viewmodel has been called (including this method).
         /// </summary>
-        public virtual async Task OnSuspendingAsync(object s, SuspendingEventArgs e) { await Task.Yield(); }
+        public virtual async Task OnSuspendingAsync(object s, SuspendingEventArgs e)
+        {
+            if (Windows.System.Profile.AnalyticsInfo.VersionInfo.DeviceFamily.Equals("Windows.Mobile"))
+            {
+                WindowWrapper.ClearNavigationServices(Window.Current);
+            }
+            await Task.Yield();
+        }
+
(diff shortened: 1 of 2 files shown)


jhalbrecht:
> Fixes #325 by clearing all cached navigation services when the app is suspending since these are re-created upon resume

Clarification for me? Is the current Cached navigation state saved, to be restored on resume?

bc3tech:
it's not the nav _state_ we're clearing here, it's the Navigation **Service** objects. Currently if you suspend and resume an app _on Mobile_, you end up with another `NavigationService` object _added_ to the collection, yet the `Bootstrapper.NavigationService` property only ever returns `.First()`... so you never get the _new_ nav service that was given to your app when it was resumed. This results in your calls to .NavigationService (to continue on past Splash) being sent to a Nav Service that isn't "attached" to the app; thereby leaving you on the Splash.

darshanrampatel:
Here's a question: how can I check that this code is included in my project?

I am using NuGet v1.1.3 of Template10, which should include this fix, but when I browse the definition of `Template10.Common.BootStrapper`, I only see:

```
        public virtual Task OnInitializeAsync(IActivatedEventArgs args);
        public virtual Task OnPrelaunchAsync(IActivatedEventArgs args, out bool continueStartup);
        public virtual void OnResuming(object s, object e, AppExecutionState previousExecutionState);
        public abstract Task OnStartAsync(StartKind startKind, IActivatedEventArgs args);
        public virtual Task OnSuspendingAsync(object s, SuspendingEventArgs e, bool prelaunch);
        public Dictionary<T, Type> PageKeys<T>() where T : struct, IConvertible;
```

I also don't have a ClearNavigationServices in `Template10.Common.WindowWrapper`

Am I missing something here?

Issue #330, opened 2015-11-12
Question marks instead of diacritics (Polish) 


arutkowski00:
Hi, I have a problem with Polish diacritical signs - instead of them diamonds with question marks are displayed. I tried clean minimal _Template10_ template with added TextBlock on MainPage - doesn't work too. In designer, diacritics are displaying properly, but not in runtime...  Without Template10, e.g. default clean UWP template, all works fine.

The text I used for testing: _Zażółć gęślą jaźń_

![przechwytywanie](https://cloud.githubusercontent.com/assets/8336970/11129784/945411aa-8981-11e5-843b-5b997a1b1d70.JPG)

arutkowski00:
I found a reason - source files have wrong default encoding. They have ASCII instead of UTF-8. After converting files from template to UTF-8 everything works fine! @JerryNixon, can you fix it?
![przechwytywanie](https://cloud.githubusercontent.com/assets/8336970/11130764/19a74246-8987-11e5-9918-b7d1274bb697.JPG)

JerryNixon:
@ruci00 Perhaps you could submit a pull request?

PR #339, opened 2015-11-15
Fixed encoding

arutkowski00:
This is a fix for issue #330 . All files with ASCII encoding were converted to UTF-8.


JerryNixon:
Thanks

PR #349, opened 2015-11-18
Add a CheckedChanged event

schuettecarsten:
Added a simple CheckedChanged event


JerryNixon:
Add event args for old and new item.

PR #356, opened 2015-11-19
improvements in ObservableItemCollection

aline-almeida:
-add new constructor
-improve ItemPropertyChanged to allow knowing which property changed
-remove handler on clearItems

### DIFF
--- a/Template10 (Library)/Common/ItemPropertyChangedEventArgs.cs
+++ b/Template10 (Library)/Common/ItemPropertyChangedEventArgs.cs
@@ -1,0 +1,16 @@
+﻿using System.ComponentModel;
+
+namespace Template10.Common
+{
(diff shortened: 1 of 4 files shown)


JerryNixon:
Why make hamburger disposable?

aline-almeida:
You're right, it doesn't have to be. I've changed that.

JerryNixon:
I think we should add ChangedItemIndex.

JerryNixon:
Do you like ItemChanged or ChangedItem?

JerryNixon:
I like ChangedItem.

aline-almeida:
ChangedItem sounds better!

JerryNixon:
Can we add ChangedItemIndex ?

aline-almeida:
I've already done that.

PR #361, opened 2015-11-20
Fixed PageParameter comparision

vinodshinde:
When page parameters are set on HamburgerMenuItem, the menu was not getting selected on startup. After debugging found out that the == operator does not work on object type and therefore replaced with Equals.


JerryNixon:
Does this check for nulls?

vinodshinde:
Good Catch. The null issue is fixed now.

PR #362, opened 2015-11-22
Invoke PropertyChanged handler from UI thread

artfuldev:
This was initially present, has been removed some time back, wondering why. In the absence of this wrapping, sometimes we get the "The application called an interface that was marshalled for a different thread" exception. Initially in the Template 10 project (when it was only Template10, Blank and Minimal, before these Library and Services projects), it was present as `(App.BootStrapper).Dispatch(() =>{});`. In the current version I have referenced the path to the `DispatcherWrapper`. Maybe we can bring an `IDispatcherWrapper` reference into `BootStrapper` itself?

### DIFF
--- a/Template10 (Library)/Common/BootStrapper.cs
+++ b/Template10 (Library)/Common/BootStrapper.cs
@@ -61,7 +61,13 @@ namespace Template10.Common
         /// it is only a helper property to provide the NavigatioNService for
         /// the first Frame ultimately aggregated in the static WindowWrapper class.
         /// </summary>
-        public Services.NavigationService.INavigationService NavigationService => WindowWrapper.ActiveWrappers.First().NavigationServices.First();
+        public Services.NavigationService.INavigationService NavigationService => WindowWrapper.Default.NavigationServices.First();
+
+        /// <summary>
+        /// This is the DispatcherWrapper of the current WindowWrapper. This is used to get
+        /// the DispatcherWrapper of the "current" UI thread.
+        /// </summary>
+        public IDispatcherWrapper Dispatcher => WindowWrapper.Current().Dispatcher;
 
         /// <summary>
         /// The SplashFactory is a Func that returns an instantiated Splash view.
--- a/Template10 (Library)/Common/WindowWrapper.cs
+++ b/Template10 (Library)/Common/WindowWrapper.cs
@@ -9,9 +9,9 @@ namespace Template10.Common
     // DOCS: https://github.com/Windows-XAML/Template10/wiki/Docs-%7C-WindowWrapper
     public class WindowWrapper
     {
-        public static WindowWrapper Default() => ActiveWrappers.FirstOrDefault();
+        public static WindowWrapper Default => ActiveWrappers.FirstOrDefault();
         public readonly static List<WindowWrapper> ActiveWrappers = new List<WindowWrapper>();
-        public static WindowWrapper Current() => ActiveWrappers.FirstOrDefault(x => x.Window == Window.Current) ?? Default();
+        public static WindowWrapper Current() => ActiveWrappers.FirstOrDefault(x => x.Window == Window.Current) ?? Default;
         public static WindowWrapper Current(Window window) => ActiveWrappers.FirstOrDefault(x => x.Window == window);
         public static WindowWrapper Current(INavigationService nav) => ActiveWrappers.FirstOrDefault(x => x.NavigationServices.Contains(nav));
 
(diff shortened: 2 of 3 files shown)


JerryNixon:
There is no reason to assume that this will work. Why? Because it is easily possible for a developer to open a view with Open() instead of Navigate(). This opens a new window, and starts a new UI thread. But Current() should return the correct one, right? Hmm. Could you verify this works on view models in a secondary window? I can't recall the full reason I removed it in the first place. I thought it was that but maybe it was something else. Anyway, would you verify before I merge this?

JerryNixon:
Actually, instead of Bootstrapper, try using WindowWrapper.Current().Dispatcher.Dispatch().

artfuldev:
I will try and update, thanks. Totally missed the Open()

artfuldev:
If WindowWrapper.Current().Dispatcher works, can we add a reference to it from BootStrapper? That way we can have BootStrapper.Current.Dispatcher (=> WindowWrapper.Current().Dispatcher;) in the code and change the implementation of BootStrapper.Dispatcher as we find it fit. Anyway I will try everything out and then update, thanks. I never have used a secondary window, so totally missed it.

artfuldev:
Also, any reason we cant have Current and Default in WindowWrapper as Calculated Properties instead of methods?

artfuldev:
WindowWrapper.Current().Dispatcher works either way. I'll update.

JerryNixon:
I will not be merging this until our next NuGet push as this is a breaking change.

artfuldev:
I believe there are 4 changes:
1. Changing Default() to Default in WindowWrapper. The only reference I found of this was in WindowWrapper.Curent() and I have changed it. This shouldn't break anything else.
2. Changing the implementation of NavigationService in BootStrapper to use WindowWrapper.Default. This just a refactoring and shouldn't affect functionality.
3. Adding an IDispatcherWrapper in BootStrapper. This is just an extension of the class and shouldn't break anything too.
4. The actual change. Invoking PropertyChanged handlers using BootStrapper.Current.Dispatcher. This is also just a refactoring and (hopefully) shouldn't break anything.

Am I missing something here? I do not have the latest codebase so I might have missed something, just making sure that I haven't unintentionally introduced an unnecessary breaking change. Because if Default() is the problem we can just undo the change and update the rest. Although I do like WindowWrapper.Default and WindowWrapper.Current more than WindowWrapper.Default() and WindowWrapper.Current() but that's just personal preference.

JerryNixon:
I have added `[Obsolete]` to the `NavigationService` property in Bootstrapper.

JerryNixon:
I will not be changing Current() to Current unless you can give me a technical reason. Why? Because Current() shows up in a lot of places, not just this class. And, for consistency, either it stays Current() or is changed to Current (or Default) in every single place. And that's why it's a breaking change.

JerryNixon:
Okay. In BindableBase, I have added to the `catch { }` of RaisePropertyChanged() a secondary attempt to Invoke() the event using `WindowWrapper.Current().DIspatcher.Dispatch();` This should default to NOT using the dispatcher, and then attempting to use the dispatcher if the UI thread is in correct and results in an exception. If both attempts fail, then it will `throw;`. This make sense?

I think this best because it allows the developer to call a property change and use any thread context they want to use. If they fail to provide a valid one, then it will attempt to auto-correct it. If we did the dispatcher call in the original `try` then the developer would not have this option.

JerryNixon:
I take that back, I have not (I have added  [Obsolete]  to the  NavigationService  property in Bootstrapper.) as this would cause App.xaml.cs to see an error when there is none.

artfuldev:
Sorry due to floods in Chennai I had no connectivity.

I have not changed Current() to Current. And for consistency it should remain Default() so we can make do with what is there right now. As I mentioned, it was just personal preference.

Calling BootStrapper from anything inside Mvvm is not an intuitive API, I agree. Finally we have arrived at a better solution than the one I tried to provide. That WindowWrapper.Current().Dispatcher.Dispatch in the catch block is something I didn't think of. Nice one, a pleasure to see this code develop :)

Yes I too wish the NavigationService was protected. Also, calling BootStrapper.Current.NavigationService returning the default navigation service instead of the 'current' (WindowWrapper.Current().NavigationServices.First()) kind of confuses me. There is a lot of tie-in to WindowWrapper with reference to navigation I suppose.

In this regard, will having a property NavigationService inside WindowWrapper like

```
public INavigationService NavigationService => NavigationServices.FirstOrDefault();
```

be more readable?
We seem to be calling WindowWrapper.Current().NavigationServices.First() a lot. Just a thought.

artfuldev:
Since we're overloading Current() it seems best to leave it that way anyway. That was my first thought.

artfuldev:
Actually I still get the Exception and crash. It is a native exception and isn't 'caught' in the catch block. Any help?

PR #381, opened 2015-12-02
Added a simple CheckedChanged event

schuettecarsten:
Added a simple CheckedChanged event and a new ValueChangedEventArgs class.


JerryNixon:
Let's call this event SelectedChanged. Remember that the newValue and oldValue could either be null.  Also change the event args to simply be ChangedEventArgs. Thank you for putting it in the Common namespace. That was the right place for it, as things are today.

schuettecarsten:
Updated cod for new class/event named. Null handling should be no problem with current implementation.

Issue #280, opened 2015-10-26
NavigationService.SaveNavigation fails on objects


schuettecarsten:
NavigationService.SaveNavigation() throws an exception when CurrentPageParam cannot be stored. Exception says that "ToString()" should be implemented. Unfortunately, this is wrong, ToString() is not called here.

```
            try { state["CurrentPageParam"] = CurrentPageParam; }
            catch
            {
                throw new Exception("Failed to serialize page parameter, override/implement ToString()");
            }
```

Also several lines later I get the exception "GetNavigationState doesn't support serialization of a parameter type which was passed to Frame.Navigate":

```
    state["NavigateState"] = FrameFacade?.GetNavigationState();
```

I think the way how Template10 saves the application state should be reworked. Currently it only works when using strings. But UWP10 platform allows to pass objects as navigation parameters.

JerryNixon:
Yeah, I hate this. When you navigate to a page, the best practice is to pass a string or integer that you use to locate the record. It is not a best practice to pass objects. Why? Because SaveNavigationState will only work with simply objects. The solution to your problem is to pass only strings/int. This is 100% not a Template 10 issue, this is just how the Frame state works.

JerryNixon:
One more note, to do what you want with state[] you can serialize it ahead of time.

JerryNixon:
Let me just vent on this a little more. Why did they make it so that you could pass objects only to make that scenario break things? There are some cases where it is important to pass an object. The first is protocol activation to return a result. This requires that the resulting page have a reference to the launch argument in order to return a value. However, I can solve this with Template 10, using our Bootstrapper.SessionState dictionary. You can, too. 
1. Bootstrapper.SessionState[key] = value
2. NavServices.Navigate(type, key)
3. param = Bootstrapper.SessionState[key]

Make sense?

schuettecarsten:
The SessionState is persisted when App is suspended?

JerryNixon:
No it is not. Which means, you would want code that, if it is not present, would load it from scratch otherwise. Make sense? It's basically saying "use this data unless it isn't there"

schuettecarsten:
Jerry, I will send you a pull request soon that will allow users to override two methods in bootstrapper/app to serialize/deserialize custom objects here.

schuettecarsten:
I've send a pull request today that adds two protected method to allow users to implement their own solution to serialize objects - if they want: https://github.com/Windows-XAML/Template10/pull/388

PR #388, opened 2015-12-04
Add custom methods to serialize/deserialize for PageParam data

schuettecarsten:
This pull requests adds two additional methods to NavigationService to serialize and deserialize CurrentPageParam object into strings. These methods only call ToString() in default implementation, if custom behavior is neccessary, you can create a descendent class of NavigationService and extend it in your project.

To make it possible to use a custom NavigationService, I've also added a simply factory method to Bootstrapper.cs which can also be overridden in custom App.xaml.cs class to create a custom NavigationService.

Third, I've added a Task.Yield().GetAwaiter().GetResult() call while wating for Frame.Content to become available on resume. There has been a while(...) {} statement, which eats up CPU while waiting, a yield is better here.

For code cleanup, I also changed several NavigationService to INavigationService in Bootstrapper and Shell.xaml.cs.


schuettecarsten:
This will also solve https://github.com/Windows-XAML/Template10/issues/280

schuettecarsten:
Additional commits to allow descendant classes of NavigationService: Make constructor "protected internal".

schuettecarsten:
Now it works :)

JerryNixon:
So, this is a breaking change since it invalidates INavigationService. I am also not 100% sure this implementation is better than requiring the developer to add it to the logic of their page instead. That being said, I am going to move this to a next version T10 milestone where we can review it and include it there. You will not see it included in this weeks NuGet update to 1.0.9.

schuettecarsten:
It should not be a breaking change, behavior of existing applications is not changed.

Please note that it is not possible to implement custom serialization behavior with the curretn implementation. Assigning an object to state["name"] throws an exception for complex object and does not call ToString(). state[] is only able to handle strings internally.

JerryNixon:
Where are there both CreateNavigationService and NavigationServiceFactory?

schuettecarsten:
Sorry, closed this accidentially.

schuettecarsten:
To answer your question: I need the method to only create the instance of NavigationService descendant class in my code. The NavigationServiceFActory method cannot be overridden because it contains too much stuff. So I moved only the "new NavigationService(..." into a separate method that can be used to create your own navigation service implementation

schuettecarsten:
Should be mergeable without conflicts

PR #392, opened 2015-12-06
Fix crash in cortana sample.

pekspro:
The Cortana sample crashed on my computer. Probably because of Cortana is not supported in my region. This change shows an error message instead.


JerryNixon:
Terrific work.

Issue #385, opened 2015-12-04
PageHeader has issues displaying secondary commands


oadugmore:
When you expand the PageHeader control, it cuts off secondary commands if they take up a certain amount of screen space. I've only been able to reproduce on Mobile. I've confirmed with an unmodified Template10 Minimal project. Easiest to reproduce if you put your device in landscape orientation.

MichaelPetrinolis:
Hi, if you replace the page header with a command bar, it has the same behavior? Page header is just a subclass of command bar.

oadugmore:
No, regular command bar works fine.

JerryNixon:
Hmm. Since `class PageHeader: CommandBar` this should not be happening.

mvermef:
I have a screen shot of this from my mobile (920 running 10586).  It occurs if you navigate from one page back to the page with the secondary command bar and open the command bar while the cursor is in the text box without closing the keyboard first. It will also occur on the first navigation to MainPage
Subsequent menu activations do not yield the truncated menu.
![wp_ss_20151204_0002](https://cloud.githubusercontent.com/assets/2018777/11598351/0609c324-9a87-11e5-9812-f9a0accb4dfb.png)

oadugmore:
Here's a screenshot from my Icon. I tried removing the Template 10 NuGet package and re-adding it but that didn't fix the problem.
![wp_ss_20151204_0001](https://cloud.githubusercontent.com/assets/13210085/11604089/3f39104e-9a9d-11e5-8e06-9a7ab99f0e65.png)

HassanElZarkawy:
I also get the same issue that @mvermef and @oadugmore get! However if you close and reopen the menu it'll work just fine! weird

callummoffat:
I am seeing this behavior as well.

MichaelPetrinolis:
Please check if [this PR](#398) fixes the issue

oadugmore:
~~#398 makes the PageHeader smoother, fixing a minor stuttering problem I was also noticing, but it does not fix the truncation issue for me.~~

**EDIT:** #398 fixes the issue

MichaelPetrinolis:
In order to check the fix, please merge the latest from upstream in your fork and then merge the PR. Then run the minimal from the solution.

oadugmore:
[...]

PR #398, opened 2015-12-08
fixes issue #385 with secondary commands

MichaelPetrinolis:
Modified the template to fix the problems discussed at #385

### DIFF
--- a/Template10 (Library)/Themes/Generic.xaml
+++ b/Template10 (Library)/Themes/Generic.xaml
@@ -568,106 +568,220 @@
                                         </Storyboard>
                                     </VisualTransition>
                                 </VisualStateGroup.Transitions>
-                                <VisualState x:Name="CompactClosed" />
+                                <VisualState x:Name="CompactClosed"/>
                                 <VisualState x:Name="CompactOpenUp">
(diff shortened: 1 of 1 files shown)


JerryNixon:
Please update all `DiscreteObjectKeyFrame` animations to VisualState.Setters/Setter instead. The reason is not just because it is more readable, but also because of performance. Spinning up a Storyboard for this type of work is needless overhead, esp when there is a lot of them.

MichaelPetrinolis:
Sure, the reason I used the storyboards was that the commandbar works as expected and the default template use storyboards. I will try to use setters instead.

MichaelPetrinolis:
It seems that using a setter for the Rect of the OverflowContentRootClip causes this behavior.
I could leave only the OverflowContentRootClip on an ObjectAnimationUsingKeyFrames, but it flickers when opening. The StoryBoard version is smoother and does not flicker.

MichaelPetrinolis:
Hi @jerrynixon , any suggestions?

JerryNixon:
Re-templating this at this level will be trouble when it is updated internally. But is good now.

PR #405, opened 2015-12-11
UWP1.1 compatibility

callummoffat:
As of the latest revision of the Template10 library, it is fully compatible with UWP1.1 (SDK build 10586). To that end, this pull request makes build 10586 the targeted SDK version. 10240 is still supported as the minimum.

### DIFF
--- a/Template10 (Library)/Template10 (Library).csproj
+++ b/Template10 (Library)/Template10 (Library).csproj
@@ -11,7 +11,7 @@
     <AssemblyName>Template10Library</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.10240.0</TargetPlatformVersion>
+    <TargetPlatformVersion>10.0.10586.0</TargetPlatformVersion>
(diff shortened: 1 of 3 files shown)


schuettecarsten:
Why is this neccessary? Will it break anything on existing applications that still require 10240 SDK? Or will this change make things possible in Template10 that are not possible now, are neccessary somewhere and will not break anything on older platforms?

According to this link, I doubt that 10586 SDK version is a valid requirement for Template10:
http://mtaulty.com/CommunityServer/blogs/mike_taultys_blog/archive/2015/12/01/windows-10-10586-visual-studio-2015-update-1-contract-definitions.aspx

An application which specifies the same version as both the minimum and the target version only needs to think about whether API contracts are present as the app crosses device families. But an application which specifies 10586 as the target version and 10240 as the minimum version has to think about whether API contracts are present and at the right version as it crosses device families.

JerryNixon:
Hey guys, let's release 1.0.9. We will branch it to preserve the 10240 support side of things, and then we will introduce version 1.10586.1. This is not a technical decision, please understand, this is a simplification decision for us and for developers trying to use T10. That being said, this is also a technical decision as we will not actually support both.

callummoffat:
OK. I'll keep working on compatibility with UWP1.1, and I'll submit a pull request post-1.0.9. Although, shouldn't the first UWP1.1 release be 1.10586.0?

PR #409, opened 2015-12-11
Unnecessary template zip duplication removed 

mickut:
The vstemplate update makes the template show in two parent categories above Template10. TargetDir2 made optional.


JerryNixon:
Okay, I am checking on this.

JerryNixon:
Terrific.

PR #416, opened 2015-12-15
Navigation suspension improvement

pekspro:
I found a few things in suspension management that affected my application. The first problem was in NavigationService where the frame state was cleared every time a navigation occurred to a new page. So states for pages in the back stack was lost if the app was suspended.
The second problem was in FrameFacade where the state was saved for each page type. So if you had several instances of the same page type only one state could be saved.

So I did a few tweaks and now it works without any problems :-). I also added a very simple sample application that I used for testing - NavigationSample. In this sample it’s possible to add several instances of the same page and the state is preserved no matter which page is visible on suspension. Back and forward navigation works as expected as well.

I also added a small extension for IDictionary<string, object> called GetOrDefault. This will return the value from the IDictionary if it exists and has the right type, otherwise a default value. Very useful when restoring state after navigation.


JerryNixon:
This level of change will take some time to review.

pekspro:
That’s good, I want this to be reviewed properly :-). The critical changes are in FrameFacade.cs and NavigationService.cs. And it’s just a few lines that has been changed. Fairly easy I think but it’s still very possible I’ve done something bad :-).

The extension is in IDictionaryExtension and the rest of the files is in the test application. None of these are important I think but very useful for testing.

JerryNixon:
There are a lot of stray extras in this PR. Let's approach this in parts. Could you submit an Issue with the requirements? I am going to close this PR, but it will stay around so we can ref it. I want to make changes like this with small PRs rather than monolithic ones. It's just too tricky to parse.

PR #418, opened 2015-12-16
Fixed null exception for WindowWrapper

calloncampbell:
Updated Minimal Template and Settings sample with code to not show the Back Button setting when the app is NOT running in Desktop Disposition.


JerryNixon:
I do not want to add more to Minimal. I have added your update to DeviceUtils to the main branch.

PR #421, opened 2015-12-18
Incorrect di, switched variables

ivesely:
windowWrapper = WindowWrapper; => WindowWrapper = windowWrapper;


JerryNixon:
Thanks for this PR. Great catch.

PR #423, opened 2015-12-19
Posibility to show backbutton without items on backstack was added.

ivesely:
I need to show back button without items on backstack.
Until now I'm adding dummy item to backstack to show shellbackbutton and delete dummy item to hide shellbackbutton.

With my change:

```
//show shell backbutton
BootStrapper.Current.ForceShowShellBackButton = true;
BootStrapper.Current.UpdateShellBackButton();

//hide shell backbutton
BootStrapper.Current.ForceShowShellBackButton = false;
BootStrapper.Current.UpdateShellBackButton();
```


JerryNixon:
Terrific idea.

PR #424, opened 2015-12-19
Add custom methods to serialize/deserialize for PageParam data

schuettecarsten:
**\* This is an updated (rebased) request for #388.

This pull requests adds two additional methods to NavigationService to serialize and deserialize CurrentPageParam object into strings. These methods only call ToString() in default implementation, if custom behavior is neccessary, you can create a descendent class of NavigationService and extend it in your project.

To make it possible to use a custom NavigationService, I've also added a simply factory method to Bootstrapper.cs which can also be overridden in custom App.xaml.cs class to create a custom NavigationService.

Third, I've added a Task.Yield().GetAwaiter().GetResult() call while wating for Frame.Content to become available on resume. There has been a while(...) {} statement, which eats up CPU while waiting, a yield is better here.

For code cleanup, I also changed several NavigationService to INavigationService in Bootstrapper and Shell.xaml.cs.

### DIFF
--- a/Samples/BottomAppBar/Views/Shell.xaml.cs
+++ b/Samples/BottomAppBar/Views/Shell.xaml.cs
@@ -22,7 +22,7 @@ namespace Sample.Views
         public static Shell Instance { get; set; }
         public static HamburgerMenu HamburgerMenu { get { return Instance.MyHamburgerMenu; } }
 
-        public Shell(NavigationService navigationService)
+        public Shell(INavigationService navigationService)
         {
             Instance = this;
             InitializeComponent();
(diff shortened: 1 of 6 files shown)


schuettecarsten:
Added a new commit to make sure that original behaviour of Template10 is not changed (by returning parameter values unchanged in default implementation) and add missing support for Navigate call

JerryNixon:
PLEASE create a navigation sample.

JerryNixon:
This needs to be demonstrated and documented.

JerryNixon:
PLEASE update Nav docs.

JustinXinLiu:
Has this been documented yet?

PR #434, opened 2015-12-24
Sharing navigation button

### LINKED ISSUE #395: Manually updating HamburgerMenu highlighted button #372
Following on from #372, I've not been able to get this to work using the IsChecked property.

Shell.xaml

```
                <Controls:HamburgerButtonInfo x:Name="MyAccountPageButton" ClearHistory="False" PageType="views:AccountPage">
                    <StackPanel Orientation="Horizontal">
                        <FontIcon Width="48" Height="48" FontFamily="Segoe MDL2 Assets" Glyph="&#xE82D;"/>
                        <TextBlock Margin="12,0,0,0" VerticalAlignment="Center" Text="Account" />
                    </StackPanel>
                </Controls:HamburgerButtonInfo>
```

Shell.xaml.cs

```
    public static Shell Instance { get; set; }
    public static HamburgerMenu HamburgerMenu { get { return Instance.MyHamburgerMenu; } set { } }
    public static HamburgerButtonInfo AccountPageButton { get { return Instance.MyAccountPageButton; } set { } }
    public void HighlightButton()
    {
        HamburgerMenu.NavAreaBackground = new Windows.UI.Xaml.Media.SolidColorBrush(Colors.Red);
        AccountPageButton.IsChecked = true;
    }
```

AnotherPage.xaml.cs

```
    public void ToggleButton() {
        Shell.Instance.HighlightButton();
    }
```

When the HighlightButton method is called, the NavAreaBackground successfully changes to red. When the IsChecked property is set to true, the active page switches back to the PageType, which takes you away from the page you are on.

I have also tried using the below

```
HamburgerMenu.PrimaryButtons[index].IsChecked = true;
```

but with no success.

mitkodi:
The same navigation button can be selected when navigating to more than one page, based on attached to the page attribute.
Resolves #395

### DIFF
--- a/Template10 (Library)/Controls/HamburgerButtonInfo.cs
+++ b/Template10 (Library)/Controls/HamburgerButtonInfo.cs
@@ -92,7 +92,14 @@ namespace Template10.Controls
             set { Set(ref _MaxWidth, value); }
         }
 
-        public override string ToString() => string.Format("{0}({1})", PageType?.ToString() ?? "null", PageParameter?.ToString() ?? "null");
+		string _navigationGroup;
+		public string NavigationGroup
+		{
+			get { return _navigationGroup; }
+			set { Set(ref _navigationGroup, value); }
+		}
+
+		public override string ToString() => string.Format("{0}({1})", PageType?.ToString() ?? "null", PageParameter?.ToString() ?? "null");
 
         public event RoutedEventHandler Selected;
         internal void RaiseSelected()
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
@@ -80,16 +80,19 @@ namespace Template10.Controls
 
         void HighlightCorrectButton(Type pageType = null, object pageParam = null)
         {
-            if (!AutoHighlightCorrectButton)
-                return;
-            pageType = pageType ?? NavigationService.CurrentPageType;
-            pageParam = pageParam ?? NavigationService.CurrentPageParam;
-            var values = _navButtons.Select(x => x.Value);
-            var button = values.FirstOrDefault(x => x.PageType == pageType 
-                && (x.PageParameter == null 
-                || x.PageParameter.Equals(pageParam)));
-            Selected = button;
-        }
+			if (!AutoHighlightCorrectButton)
+				return;
+			pageType = pageType ?? NavigationService.CurrentPageType;
+			var navGroup = pageType.GetTypeInfo().GetCustomAttribute<NavigationGroupAttribute>();
+			pageParam = pageParam ?? NavigationService.CurrentPageParam;
+			var values = _navButtons.Select(x => x.Value);
+			var button = values.FirstOrDefault(x => x.PageType == pageType
+				&& (x.PageParameter == null
+				|| x.PageParameter.Equals(pageParam)));
+			if (button == null && navGroup != null)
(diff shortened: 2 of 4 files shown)


Template10 (Library)/Controls/HamburgerButtonInfo.cs
```diff
@@ -92,7 +92,14 @@ public double MaxWidth
             set { Set(ref _MaxWidth, value); }
         }
 
-        public override string ToString() => string.Format("{0}({1})", PageType?.ToString() ?? "null", PageParameter?.ToString() ?? "null");
+		string _navigationGroup;
```
chrisleyva, line 95: Sorry to come across as tedious, but why are all these extra spaces added? Is it just an issue with github's Diff Viewer? Looking at the rest of your changes, it seems your Tools > Options > Text Editor > Tab Spaces are set to something non-standard. I think it should be 4 by default.

JerryNixon:
I like this solution, but feel it is overly complex for developers to implement. We need to brainstorm if we can get IsChecked to work. I know there's a way, and if it requires us to change the pipeline so that HighlightCorrectButton gets called in a different order, that would be worth it.

Template10 (Library)/Controls/HamburgerButtonInfo.cs
mitkodi, line 95: I think the difference comes from "Keep tabs" which is true in my settings, not from tab size (which is 4).

mitkodi:
If we want to select a button in code without navigating to the page related to that button we can:
- set IsChecked or Selected in Page.OnNavigatedTo (relaying on that that the new navigation will not happen if the current is not finished yet).
- set IsChecked or Selected from anywhere, but it will require small changes in HamburgerMenu to conditionally skip navigation in HamburgerMenu.SetSelected. In this case, I think that using HamburgerMenu will require less changes than using HamburgerButtonInfo because it doesn't have reference to HamburgerMenu where the navigation happens.

JerryNixon:
Just remove PageType from the Button.

darshanrampatel:
@mitkodi, do you have an example as to how to achieve the below

> set IsChecked or Selected in Page.OnNavigatedTo (relaying on that that the new navigation will not happen if the current is not finished yet).

I can't use Selected any more as it's been made an internal property.

What I need is to navigate to a new page (that doesn't have its own button), but manually highlight an existing button.

mitkodi:
I think that HamburgerMenu.Selected is still public, so you should be able to use it.
Btw, I had an idea (#395) to replace using of Selected property with a Select method which will help to select a button not only from Page.OnNavigatedTo but from anywhere, but not sure if anyone else is interested in it.

darshanrampatel:
@mitkodi, yes, the idea of having a Select method sounds good, if you don't mind sharing what code you have so far.

mitkodi:
I'll made a PR (maybe), but it could be rejected so at the moment I think it's better to relay on what we already have:
1. Create a method in your **Shell** class for selecting the button you want to select in code. In it just set Selected of your HamburgerMenu to the desired HamburgerButtonInfo instance.
2. In **OnNavigatedTo** of your page call that method via **Shell.Instance**.

darshanrampatel:
@mitkodi, thanks. I've got this to work using the method you suggested and by removing **PageType** from the buttons.

mitkodi:
Maybe something was changed that requires to remove PageType from the button. I think it's a restriction that should be overcome, so I'm thinking to look for solution...

PR #443, opened 2016-01-02
Fixes HamburgerMenu flicker.

### LINKED ISSUE #410: Hamburger Menu doesn't stay open on first click
Using the Template 10 (Minimal) project, the Hamburger Menu will not stay open the first time you click it. It will open and close instantly, next time you click the Hamburger Menu it stays open as it's supposed to. 

Edit: Forgot to mention that this only applies to VisualStateNarrow.

chrisleyva:
Fixes #410. The HamburgerMenu would flicker if the app was started in the Narrow VisualState and AutoHighlightCorrectButton was true.


JerryNixon:
Is that the correct place to set Loaded=true?

chrisleyva:
You're right, that is setting _navButtonsLoaded to true for each button in the hamburger menu. Rookie mistake :-) I was testing with a single button in the HamburgerMenu.

I'll take another look at it soon. Good catch.

PR #445, opened 2016-01-04
Add PageHeaderSample.

pekspro:
I created a simple demo application to myself to figure how to work with the PageHeader when items are selected in a list. I created something that is quite similar to the behavior in the Photo-app. There are two page headers, one for normal mode and another in selection mode. Is this a sample you want? It’s not complicated but it took a while to figure out :-).

The interesting stuff is in MainPageViewModel.cs, MainPage.xaml and MainPage.xaml.cs.


JerryNixon:
Thank you.

PR #446, opened 2016-01-05
Added TextBoxCharactersEnteredBehavior

artfuldev:
This behavior triggers its actions when a specified number of characters is entered in the attached TextBox control.
Useful for scenarios like automatic validation of verification code when the n-digit code is entered.


artfuldev:
Adding an InvokeCommandAction or CallMethodAction to handle is easy for both MVVM and non-MVVM solutions

artfuldev:
Can be used as such:

```
<interactivity:Interaction.Behaviors>
    <behaviors:TextBoxCharactersEnteredBehavior CharacterCount="6">
        <core:InvokeCommandAction Command="{Binding VerifyCommand}" />
    </behaviors:TextBoxCharactersEnteredBehavior>
</interactivity:Interaction.Behaviors>
```

JerryNixon:
I am not sure of the use case for this. Could you tell me one?

artfuldev:
Say we use an OTP (One Time Pin) for verification of an account/user.
We send the user an SMS to verify his mobile, or an email to verify his email address, and wait with an input prompt for him to enter. On the phone, if we know the number of digits/characters an OTP has, we don't even have to wait for him to press the Submit button, we can automatically start processing his input the moment he enters the n characters into the TextBox.

Or take an example of a TextBox expecting a credit card number to be input.
1. If there is one text box, as soon as the 16 characters/digits are entered, it can be processed to see if it is a valid credit card, without any interaction from the user like moving focus to another control, or clicking a button, or moving focus out of this control. Or, for every 4 characters entered, a space can be automatically added in the textbox, for readability.
2. If there are a group of 4 text boxes, each expecting 4 characters (digits), as soon as 4 characters are entered in the one, the focus can be moved to the subsequent textbox, whereby for the user it is a seamless experience of typing in a full number. Also, once the last textbox is filled with the 4 characters (digits), the focus can be moved to the "date" field while simultaneously validating the credit card.

I think the credit card example might be more familiar.

artfuldev:
For a visual (here the card type is checked upon entering the first few characters, a space is added after every 4 characters, and a validation is done upon entering the full number):

Payment GIF

JerryNixon:
Very nice explanation. Thank you. But this behavior is already possible simply by using the ConditionalAction already in T10. Set the conditional action to TextBox.Text.Length/EqualTo/4 with a nested InvokeAction. This may not be a necessary new behavior to add. Do you think?

artfuldev:
Um I have to admit that I wasn't aware of Condition Action's several usages. I have so much to learn...

The solution you are referring to uses EventTriggerBehavior with EventName="TextChanged" with ConditionalAction right? I just read the Wiki entry on Conditional Action and this is what I could come up with.

I think this behavior more clearly explains the case, but it's your call whether it should be added as a separate behavior to Template10. It's just one for convenience I guess :)

artfuldev:
A realeted question, why can't Behaviors and Actions rest in separate namespaces?

JerryNixon:
Perhaps we will rev T10 and change some namespaces. But for the sake of SIMPLE it's an easy single place for developers using T10 to find behaviors and actions (which are brothers after all).

artfuldev:
Is SIMPLE some acronym that I don't know, or did you just mean simplicity? Please illuminate

PR #455, opened 2016-01-06
Page header sample

pekspro:
I did two minor things in this pull request:
- Added PageHeaderSample to the solution file (was lost somehow).
- Simplified some async code.


JerryNixon:
Could you rebase this with the current source? Changes are great.

pekspro:
Sure. My latest two commits should fix this. But I’m a bit puzzled by GitHub right now :-). It looks like I can’t add these commits to this pull request. Is this something you are allowed to do?

Otherwise I close this and create a new pull requests. But I’m curious about how this works :-).

PR #450, opened 2016-01-06
Page header sample Fix

artfuldev:
[A commit](https://github.com/Windows-XAML/Template10/commit/1d275270a8d6db54b4a9e5b82da1423d8e1128ca#diff-8bbb2d0d4796bc83bf559387de4db2e4) from #424 broke the sample, because Shell was constructed using NavigationService created from NavigationServiceFactory. Whereas now the NavigationServiceFactory returns an INavigationService, the Shell constructor expects a NavigationService. So changed constructor of shell to expect INavigationService instead.

PR #458, opened 2016-01-07
Updated PageHeaderSample

lombo75:
PageHeaderSample project wouldn't build unless changing it


artfuldev:
The original author has added another pull request which also solves this #455

lombo75:
Ok, thanks didn't realize

artfuldev:
Don't worry, I did the same with #450 LOL

PR #465, opened 2016-01-08
Add support for very large NavigationState

schuettecarsten:
On one of our applications we have a very large NavigationState which is larger than 8000 bytes. Unfortunately, current implementation uses ApplicationDataContainer which supports only 8000 bytes per value.

This pull requests add two new methods to split large NavigationState/CurrentPageParam values into chunks of 8k and put them into an ApplicationDataCompositeValue which is then stored in the ApplicationDataContainer. This allows NavigationState size up to 64kb - which should be enough.

If value size is smaller than 8000 bytes, the it is stored directly. So original behaviour is unchanged.

### DIFF
--- a/Template10 (Library)/Services/NavigationService/NavigationService.cs
+++ b/Template10 (Library)/Services/NavigationService/NavigationService.cs
@@ -13,6 +13,9 @@ using Windows.UI.Xaml.Navigation;
 
 namespace Template10.Services.NavigationService
 {
+    using System.Text;
+    using Windows.Foundation.Collections;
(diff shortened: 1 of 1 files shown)


saguiitay:
Shouldn't such a code go into some form of serialization service?

schuettecarsten:
Maybe, but how the data is split into chunks is independent how the data gets serialized. It's only a solution to store strings longer than 8k in application data containers.

schuettecarsten:
Will be reworked that it uses the SettingsService, which already supports splitting long values.

PR #467, opened 2016-01-08
Added init code for AppInsights

luiscantero:


JerryNixon:
Thank you. But this is only relevant if the developer checks to include AppInsights. I think in those cases, they will be responsible to add it. We will not include it bydefault.

luiscantero:
To use AppInsights, there are four requirements:
1. AppInsights NuGet packages (included in the minimal Template10 template)
2. ApplicationInsights.config file (included in the minimal Template10 template)
3. Internet capability (included in the minimal Template10 template)
4. Init code in App.xaml.cs  (NOT included in the minimal Template10 template)

It seems kind of random that one step is missing. The dev might be let to think that the init code is located in the library code, since the rest of the requirements are present.

Also, this init code is even included in the Univ Apps Blank template. It doesn't have any side-effects if the instrumentation key is missing (thus disabling AppInsights by default).

Thanks.

PR #471, opened 2016-01-09
Prevent duplicate key exception in HamburgerMenu

schuettecarsten:
Prevent duplicate key exception in HamburgerMenu


a-iafrate:
this solve the same problem for me!

JerryNixon:
I do not believe this is a real scenario.

schuettecarsten:
It is on our project. I do not know why, but the exception occurs sometimes when going "back".

PR #476, opened 2016-01-10
Revert changes in UpdateFullScreen to make adaptive triggers work again

schuettecarsten:


schuettecarsten:
Works now. Sometimes solutions are so simple. Thanks!

PR #478, opened 2016-01-10
Cancelable navigation

ivesely:
Added:
- \+ NavigateAsync 
- \+ OnNavigatingAsync

2 problems here:
- NavButtonBehavior GoBack does not fire BackRequested, same problem for GoForward => OnNavigatingAsync does not fire. 
- HamburgerMenu selected is not correctly updated when navigation is canceled


JerryNixon:
Okay, first thing to note, I have (just now because of you) added to NavigationService

``` csharp
public static INavigationService GetForFrame(Frame frame)
{
    return WindowWrapper.ActiveWrappers.SelectMany(x => x.NavigationServices).FirstOrDefault(x => x.Frame.Equals(frame));
}
```

That method will help me get the NavService based on the Frame in NavButtonBehavior. This make sense so far? I had not thought of that problem until now. I have updated the behavior, too. This is a very good change. Thank you for pointing it out.

JerryNixon:
I think I am in love with this PR. But I also think it needs to wait until we update for 10586 update because I would like to go one step further and delete the OnNavigatingFrom and replace it with OnNavigatingFromAsync(). Could you rebase your PR?

callummoffat:
I think this PR is awesome. Although, Jerry, T10 HEAD is currently compatible with 10586. The latest version of my production app, Petrolhead, uses it with no adverse effects whatsoever.

ivesely:
No problem

schuettecarsten:
ivesely, I think you should add a new PR with the rebased branch.

PR #483, opened 2016-01-11
Add new methods to Bootstrapper

schuettecarsten:
This pull request adds three new methods to Bootstrapper.

**OnResumingAsync:**
Same as OnResuming, but async to allow users to call and await other async methods. The old method OnResuming has been marked as Obsolete but is still called.

**OnRestoredNavigationAsync:**
This method is invoked when the navigation service restored the navigation. This method is normally called when app is restored from resume and resume+terminated state.

**OnSaveNavigationAsync:**
This method is invoked immediately before the navigation service saves the navigation state. It allows "last minute" changes on the navigation state.

### DIFF
--- a/Template10 (Library)/Common/BootStrapper.cs
+++ b/Template10 (Library)/Common/BootStrapper.cs
@@ -44,7 +44,13 @@ namespace Template10.Common
         protected BootStrapper()
         {
             Current = this;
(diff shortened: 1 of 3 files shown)


JerryNixon:
These are not necessary.

schuettecarsten:
Correct. Will send a new (small) PR for the new event in the navigation service.

PR #497, opened 2016-01-13
Add IsHardwareButtonsApiPresent property to new class ApiUtils

schuettecarsten:
Add IsHardwareButtonsApiPresent property to DeviceUtils (as static property because it will not change)


schuettecarsten:
Will send a new PR that contains the changes in only one commit

PR #521, opened 2016-01-15
Cancelable navigation

ivesely:


JerryNixon:
Looks like we were both working on this at the same time! 
See  https://github.com/Windows-XAML/Template10/commit/28a003670a2b535bcdab494fb9d96f5532430009

PR #523, opened 2016-01-16
Added CustomAutomationPeer and Accessibility attached properties

lombo75:
this PR is mainly demonstrating the difficulties that are going to occur in making accessibility work with the way the Radiobutton menu is designed. 
From my understanding so far, a Custom automation peer class will be needed with the correct provider and methods set for each control, ie. IInvoker, ISelectionItem providers

I am a newbie in WPF and accessibility so hopefully somebody else can help get this working properly, but it is a start. When testing with narrator the menu 
button works with the default automation providers for a button, with the Automation Properties set in xaml. To test with customautomationpeer class you need to uncomment the oncreateautomationpeer method, this method fires automatically when narrator is started. I suggest testing both ways to see the differences and the problems when tabbing, through the controls, also the event handlers need to be hooked up to. Each control needs automation.ID and at least Automation.Name to be seen by Narrator for which the radiobuttons/Toggle buttons don't have but can be added in code while loading. there are tools in the windows sdk for UIAutomation testing, Appscope.exe and Inspect.exe, to show which UIElements don't have values and which providers are supported by each control, I added as external tools in VS for convenience.
Probably the best Sample project to test against is the Search Sample as it has a lot of controls. Also Windows + z didn't work for me. When test with narrator to select the menu button use caps lock and spacebar, if that doesn't work just spacebar selects the button and opens menu, you can look at all commands in narrator settings. Good luck will keep persisting with this but I have to learn other parts of app to go further with it. hope it helps get things started with this enhancement.


JerryNixon:
I don't understand the intent of this PR. 
Is it to accomplish something or `demonstrating the difficulties`?

lombo75:
Both, 
1. it fixes the automationproperties for the Hamburgerbutton, if you remove the automation properties from HamburgerMenu.xaml on hamburger button you will see that narrator just calls it button as no automation properties are set, so it accomplishes adding accessibility to the hamburger menu button, but,
1. Demonstrates the difficulties 
   a) to add automation properties to the Radio/Buttons in menu (each HamburgerButtoninfo) as it is a group of radiobuttons in a scrollviewer and stackpanel, Each control needs a Name to be detected by Narrator (how to set for each Hamburgerbuttoninfo?)

a custom automationpeer is needed to set up custom pattern/provider for these controls from my understanding, otherwise it uses the default RadioButtonAutomationpeer, which is not what is required as on the surface it is not a radiobutton, it is a button control which has anything the developer wants to put inside it, 

below are provider methods and required control patterns, you will see that button uses IInvokeProvider and radiobuttons use ISelectionItemProvider, so which way to do it?
https://msdn.microsoft.com/en-us/library/windows/desktop/ee671641(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/ee671619(v=vs.85).aspx

The CustomControlAutomationPeer class just shows the methods that need to be overriden to set up the customAutomationPeer. The code doesn't affect the app unless you uncomment the onCreateautomationPeer method and turn on Narrator, but I did PR to show how it is done and hopefully get somebody that knows a bit more about UIAutomation to get it working correctly, if you don't want to merge, I understand.

I hope this helps in some way!

lombo75:
Maybe this will help to understand

https://msdn.microsoft.com/en-us/library/windows/apps/xaml/mt244349.aspx

lombo75:
@JerryNixon , the latest commit fixed narrator saying the right text for each button but I couldn't work out how to make it say button instead of radiobutton after the name of the button. The customautomationpeer might not be neccessary, will continue to work on it. The other issue is when tabbing it is still cycling through all controls in the group, so you have to tab four times to go to next button, any ideas? Tabindex property?

lombo75:
@JerryNixon , the radiobuttons and menu need to be changed to listbox or another control to get full accessibility as per the windows-universal-samples-master demos, I suppose that is why they used a listbox for the scenarios

Example Narrator knows a list has a certain amount of items in a listbox and says 1 of 4, 2 of 4 etc..., with the current design (radiobuttons) there is no way the disabled user will know how many items are in the menu or whether they are still in the menu on the next tab. 

I don't believe in wasting time on something that in the end cannot be done, unless there is something that you can think of to make this design fully accessible.

I will give it a couple of days for your response otherwise will close PR.

PR #529, opened 2016-01-16
Add serialization service to fix problems when passing complex objects as navigation parameters

schuettecarsten:
Add serialization service to fix problems when passing complex objects as navigation parameters


JerryNixon:
This REALLY needs a sample project, Carsten

PR #531, opened 2016-01-17
Optimizations and bugfixes in SerializationServices

schuettecarsten:
Minor optimizations and bugfixes in SerializationServices when dealing with null values.


schuettecarsten:
Will be re-added in a more simpler PR.

PR #535, opened 2016-01-18
Block PropertyChanged events on outdated ViewModels

schuettecarsten:
Block PropertyChanged events on viewmodels that have been "navigated from".

In our app several tasks are started to load data from a server API and write the results into the view model. When user clicks on the back button and Template10 navigates to a different view, we sometimes can see strange crashes of our app - caused by the background tasks trying to write into view model properties that are bound to the view but where the view is no longer valid.

The changes from this commit will add two properties to the view model if the view has been "navigated to" and "navigated from". If the viewmodels "NavigatedFrom" method was called, the flag is set and every RaisePropertyChanged call will no longer raise the event.


schuettecarsten:
As discussed with jerrynixon, it's better to solve this very specific problem in a custom ViewModelBase class.

PR #553, opened 2016-01-21
Fixed #551: Deserialize the navigation parameter.

### LINKED ISSUE #551: Minimal Template: Settings navigation parameter isn't handled properly anymore
The Minimal template SettingsPage.xaml.cs uses Page.OnNavigatedTo to get the pivot index. It appears that 1.1.0 changes have caused the parameter to be serialized when passed to the Page.OnNavigatedTo parameter value. This means the parameter will never be parse correctly. Jerry noted in #549 that you can use the following to handle the parameter:

``` C#
var p = Template10.Services.SerializationService.SerializationService
    .Json.Deserialize<int>(e.Parameter?.ToString());
```

craigktreasure:
This fixes #551 by deserializing the navigation parameter in the Page.OnNavigatedTo method.

### DIFF
--- a/Templates (Project)/Minimal/Views/SettingsPage.xaml.cs
+++ b/Templates (Project)/Minimal/Views/SettingsPage.xaml.cs
@@ -3,18 +3,18 @@ using Windows.UI.Xaml.Navigation;
 
 namespace Sample.Views
 {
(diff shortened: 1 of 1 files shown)


JerryNixon:
Great job.

Issue #545, opened 2016-01-21
StringFormatConverter is not supporting custom language in Template10library


ganbayaru:
I'm using your template10 library. Currently StringFormatConverter is not supporting my own language. Can you implement this one, Please?

JerryNixon:
Tell me how to do it, and I will. Or, do you want to submit a PR?

ganbayaru:
I'm using your convert function with below change. It is nice to me if it is implemented in template10Library

public object Convert(object value, Type targetType, object parameter, string language)
        {
            CultureInfo culture = new CultureInfo(language);
            string format = (parameter as string) ?? Format;
            if (format == null)
                return value;
            return string.Format(culture, format, value);
        }

JerryNixon:
Terrific upgrade

PR #568, opened 2016-01-24
StringConverter: Use CurrentCulture if language is not specified

robertmuehsig:
Current problem - introduced [here](https://github.com/Windows-XAML/Template10/issues/545)

Simple binding with the converter - without given any language:

```
            <TextBlock Margin="0,8,0,8" HorizontalAlignment="Center" FontSize="42" FontWeight="Bold" Text="{Binding MoneyBurndownCounter, Converter={StaticResource StringFormatConverter}, ConverterParameter='{}{0:C}'}" />
```

Inside the converter:

format: "{0:C}"
culture: ""
value: "0"

False behavior (currently)

"¤0.00"

Fixed behavior

"$0.00" (depending on the CurrentCulture of the system.

### DIFF
--- a/Template10 (Library)/Converters/StringFormatConverter.cs
+++ b/Template10 (Library)/Converters/StringFormatConverter.cs
@@ -11,6 +11,12 @@ namespace Template10.Converters
             var format = (parameter as string) ?? Format;
             if (format == null)
                 return value;
+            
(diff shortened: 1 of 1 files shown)


JerryNixon:
I was about to fix this myself. Thank you @robertmuehsig

PR #571, opened 2016-01-25
Add JsonSerializerSettings to JsonSerializationService

schuettecarsten:
Add JsonSerializerSettings to JsonSerializationService and allow user to modify settings


JerryNixon:
This is a good addition.

Issue #588, opened 2016-01-29
NavigationMode arg in OnNavigatingFromAsync is always New due to a race condition


craigktreasure:
The FrameFacade resets the value of the NavigationModeHint before the NavigationService is notified of the navigation. The NavigationService is referring to the NavigationModeHint after it has changed to construct the arguments. The correct value is in the event arguments. Fix should be simple. I'll send a PR.

JerryNixon:
Could you verify you are using 1.1.3?

craigktreasure:
@JerryNixon From what i see, there is still no 1.1.3. The latest release is 1.1.2. :-) https://www.nuget.org/packages/Template10/

PR #589, opened 2016-01-29
Resolving #588

craigktreasure:
The FrameFacade resets the value of the NavigationModeHint before the NavigationService is notified of the navigation. The NavigationService is referring to the NavigationModeHint after it has changed to construct the arguments. The correct value is in the event arguments.

### DIFF
--- a/Template10 (Library)/Services/NavigationService/NavigationService.cs
+++ b/Template10 (Library)/Services/NavigationService/NavigationService.cs
@@ -47,7 +47,7 @@ namespace Template10.Services.NavigationService
                     return;
 
                 // allow the viewmodel to cancel navigation
-                e.Cancel = !(await NavigatingFromAsync(false));
+                e.Cancel = !(await NavigatingFromAsync(false, e.NavigationMode));
                 if (!e.Cancel)
                 {
                     await NavigateFromAsync(false);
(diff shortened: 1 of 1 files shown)


JerryNixon:
Verify for me that when you click Back, the On navigated to mode is Back. (ensure the page you are going back to has NavigationCacheMode to Enabled)

craigktreasure:
@JerryNixon Sure. Using the Search (and Login) Sample app and with the MainPage and DetailPage configured with NavigationCacheMode Enabled, clicking back on the DetailPage results in the DetailPageViewModel.OnNavigatingFromAsync receiving a New mode and the MainPageViewModel.OnNavigatedToAsync receiving Back. The behavior is the same with NavigationCacheMode Disabled.

With this change, the DetailPageViewModel.OnNavigatingFromAsync receives a Back mode.

Unless there's something I'm missing, the issue seems pretty obvious when you look at FrameFacade.FacadeNavigatedEventHandler and FrameFacade.FacadeNavigatingCancelEventHandler. NavigationModeHint is reset to New after setting up the arguments. The value will always be New unless you get to it within a very short window. It would probably be a good idea to make NavigationModeHint private, but that would technically be a breaking change for anyone who may be referencing it.

The other option would be to move the NavigationModeHint reset after the event handlers are called.

Issue #587, opened 2016-01-29
Suggestion: Enable Navigation Service to cope with wrapped Viewmodels


ghost:
I have a project where I'm using awrapper around my ViewModels.  This is currently not supported by the NavigationService. 

I suggest adding an override to the code extracting the DataContext. 

More specifically.
These three NavigationServices methods break when the ViewModel is wrapped:
-   NavigateToAsync
-   NavigateFromAsync
-   NavigatingFromAsync
  … i.e. “page.DataContext as INavigable” returns null. 

To enable a work-around, add this function Navigation Services:

`protected virtual INavigable ExtractNavigable(Page page)
        {
            return page.DataContext as INavigable;
        }`

And replace the three instances of `page.DataContext as INavigable` with `ExtractNavigable(page)`.

With those few changes it will be possible to extend the NavigationService like this:

`public class NavigationServiceEx : NavigationService, INavigationService
    {
        public NavigationServiceEx(Frame frame) : base(frame)
        {
        }

```
    protected override INavigable ExtractNavigable(Page page)
    {
        return ForView.Unwrap<dynamic>(page.DataContext) as INavigable;
    }
}
```

`
... and 

`public class BootStrapperEx : BootStrapper
    {
        public override async Task OnStartAsync(StartKind startKind, IActivatedEventArgs args)
        {
            await Task.CompletedTask;
        }

```
    protected override INavigationService CreateNavigationService(Frame frame)
    {
        return new Services.NavigationService.NavigationServiceEx(frame);
    }
} `
```

JerryNixon:
Yikes, that's a complicated addition. Your wrapper can't simply implement INav and pass along?

ghost:
No, the challenge is that the wrapper is part of another Nuget. In my case it's Assisticant, which I really like working with. But I suspect that there are other scenarioes where ViewModel/DataContext is wrapped too for good (or bad) reasons.

Regarding complexity. I don't see it that way and all that is added to the NavigationService implementing is a method and the ability to override it. Why is that complex?

To clarify: The class NavigationServiceEx and BootStrapperEx are just for refrence and not part of the suggested change, but just an example of how to utilize the proposed "virtual INavigable ExtractNavigable" when and if needed. I've tested it and it works fine.

artfuldev:
BindableBase kind of already provides what Assisticant does, to an extent, from the looks of the home page. However, may be it does much much more...

1iveowl:
[...]

PR #590, opened 2016-01-29
Adding DataContext Unwrap Extensibility

1iveowl:
Enhancement: Enable Navigation Service to cope with wrapped Viewmodels #587

### DIFF
--- a/Template10 (Library)/Services/NavigationService/NavigationService.cs
+++ b/Template10 (Library)/Services/NavigationService/NavigationService.cs
@@ -63,6 +63,12 @@ namespace Template10.Services.NavigationService
             };
         }
 
+        // can be overriden in case the ViewModel is wrapped
+        protected virtual INavigable UnwrapNavigableFromDataContext(Page page)
+        {
+            return page.DataContext as INavigable;
+        }
+
         // before navigate (cancellable) 
         async Task<bool> NavigatingFromAsync(bool suspending)
         {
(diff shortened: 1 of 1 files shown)


JerryNixon:
I think the same result can be accomplished using ResolveForPage(). No?

1iveowl:
Not sure. I have to admit I don't know how to utilize ReSolveForPage(). Is there an sample I can look at somewhere? I'm using MVVM and IOC for my viewmodels.

JerryNixon:
Just look what it does.

``` csharp
// Bootstrapper.cs
public virtual Services.NavigationService.INavigable ResolveForPage(Type page, NavigationService navigationService) => null;

// Navigationservice.cs (NavigateToAsync)
if (page.DataContext == null)
{
    // to support dependency injection, but keeping it optional.
    var viewmodel = BootStrapper.Current.ResolveForPage(page.GetType(), this);
    if (viewmodel != null)
        page.DataContext = viewmodel;
}
```

Accomplishes the same thing you are wanting.

1iveowl:
Thanks Jerry. I've had some time to look at it and ResolveForPage is only called if the DataContext is null. When using IoC and a wrapped ViewModel, then DataContext is not null. 

What if we change the if-statement to?:

`if (!(page.DataContext is INavigable))`

Then I could override the ResolveForPage, which I think is what you are suggesting and would makes sense as well as be a move elegant solution. 

What do you think?

JerryNixon:
Crazy. Okay. I changed it to `if (!(page.DataContext is INavigable) | page.DataContext == null)`. It's one of those changes that nobody will notice but you, dude!

1iveowl:
Thank you Jerry. I appreciate it. 

I made a pull request as the current version only covers NavigationTo and not NavigationFrom. I've added code to cover this. Also, I'm raising an event in the case that the Resolver is unable to find a ViewModel. Not sure if that is how you would like it to be? I'm flexible on that one :-)

Regarding others needing this you might be right. Only, time will tell, byt to me the new code looks nicer and more precise now :-)

PR #605, opened 2016-02-02
Added Voice and Ink to TextBox Samples Project

gsantopaolo:
New Pull created
It is all about making user’s experiences on Windows 10 more personal on any device.

**InkToTextBox**
 Actually there is no simple way to fill a textbox directly with your ink.
 With this control, users will be able to ink directly on a textbox. Pain and simple you ink and textbox is filled with what you typed above. 

**VoiceToTextBox**
 Actually there is no simple way (except that for Windows 10 Phone) to fill a textbox directly with your voice.
 With this control, users will be able to click the mic button and speech. Its own voice will be automatically recognized and speech translated to text.
 Todo: improve control’s theme to make it appear more similar to the Cortana (Ask me anything one) in terms of UI and also visual feedback in terms of sounds like the real Cortana textbox one.


gsantopaolo:
this pull doesn't have conflict...

PR #619, opened 2016-02-04
small correction to T10_PropertySetting.snippet 

MauRiEEZZZ:
I think line 41 of T10_PropertySetting.snippet should be:
 get { return $helper$.Read<$type$>(nameof($name$), default($type$)); } 
 instead of 
 get { return $helper$.Read<$type$>(nameof($name$), default()); }

### DIFF
--- a/Snippets/T10_PropertySetting.snippet
+++ b/Snippets/T10_PropertySetting.snippet
@@ -38,7 +38,7 @@
       <Code Language="csharp">
         <![CDATA[public $type$ $name$
         {
-            get { return $helper$.Read<$type$>(nameof($name$), default()); }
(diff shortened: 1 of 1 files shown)


JerryNixon:
Cool. What is your opinion about removing the T10_ prefix?

MauRiEEZZZ:
As I am a beginning developer and not so sure about what goes where and how it is called.
I love the T10_ prefix, it gives me a list of everything there is to help me while maintaining the thinking pattern of using the template. So I vote for keep the prefix.

Also, I'm learning about the Mvvm pattern I hoping to propose an addition to the snippets that helps building new viewmodels. I'm not sure but I think it would be a nice addition.

PR #622, opened 2016-02-04
Adding support NavigateFrom and NavigatingFrom 

1iveowl:
This pull request extends the support to also encompase NavigateFrom and NavigatingFrom. 

Also, the overload of ResolveForPage is necessary to add a DataContext unwrapping to ResolveForPage. This existing ResolveForPage is only called with the page Type, while the new overload of ResolveForPage includes the actual Page object necessary when doing an unwrap.


JerryNixon:
Not sure why you are modifying the snippets behavior. Also, this code will duplicate the work already in ResolveForPage. I do not see how added complexity to do the same thing is worth it.

1iveowl:
Jerry it is clear to me that I have been unable to explain that this code is doing. Either that  or that you do not read my comments or code???

Once more: The current ResolveForPage will NOT provide what is needed to unwrap DataContext. In the current master only the **Type** of the page is passed to the ResolveForPage. To unwrap the viewModel the actual **Page object** is needed. Just as I wrote above. 

Without the changes I've proposed Template 10 will not have the ability to manage wrapped ViewModels. You might decide that you don't care, in which cae I will be unable to use Template10. But I'd much rather that you decided to add this capability as it hurts no one.

Regarding the snipit just ingore it, but I've changed it because I'm unable to compile the code unless I remove the snippets. Not sure why, but it is irrelevant in this context.

Issue #628, opened 2016-02-07
Navigation parameter gets overriden at app start if AutoHighlightCorrectButton = true


MatthiasNistl:
I'm using voice commands to start my app and pass different navigation parameters to the navigation service, depending on what the user said. Unfortunately, OnNavigatedToAsync of the corresponding page is fired twice, one time with the correct parameter and the second time with a parameter of null.
Steps to reproduce this in the Hamburger template: Simply add a parameter to OnStartAsync.

```
public override Task OnStartAsync(StartKind startKind, IActivatedEventArgs args)
{
    NavigationService.Navigate(typeof (Views.MainPage), "Test");
    return Task.CompletedTask;
}
```

The parameter "Test" is passed to OnNavigatedToAsync the first time, but then OnNavigatedToAsync is fired again with a parameter of null. This happens when the HamburgerMenu refreshs its buttons and only if AutoHighlightCorrectButton is set to true.

By the way, AutoHighlightCorrectButton is currently always returning true:

```
public bool AutoHighlightCorrectButton
{
    get { return true; /*(bool)GetValue(AutoHighlightCorrectButtonProperty);*/ }
    set { SetValue(AutoHighlightCorrectButtonProperty, value); }
}
```

Temporarily disabling AutoHighlightCorrectButton and re-enabling it after the first navigation should do the trick, but is there a more elegant solution?

JerryNixon:
I look forward to your Pull Request.

MatthiasNistl:
Fixed by Pull Request #633.

PR #633, opened 2016-02-08
HamburgerMenu should only navigate when all NavButtons have been loaded

MatthiasNistl:
Issue #628
Every time a NavigationButton in HamburgerMenu has been loaded "HighlightCorrectButton" is called. Within this method "SetSelected" is called with the correct button, which then fires NavigationService.Navigate with the corresponding page as parameter.
This causes a problem at app start when a parameter is passed to the first page:

```
public override Task OnStartAsync(StartKind startKind, IActivatedEventArgs args)
{
    NavigationService.Navigate(typeof (Views.MainPage), "Test");
    return Task.CompletedTask;
}
```

The first navigation works fine. But after OnStartAsync the HamburgerMenu gets loaded and "HighlightCorrectButton" gets fired, which then causes another navigation to the same page, but with a parameter of null.

To fix this the HamburgerMenu should only navigate once all NavigationButtons have been loaded. If not all buttons have been loaded the correct button will still get highlighted, but no additional navigation will be fired.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
@@ -84,6 +84,9 @@ namespace Template10.Controls
                     if (!any)
                         // this is the default color if the user supplies none
                         AccentColor = Colors.DarkOrange;
+
+                    if (NavButtonCount == 0)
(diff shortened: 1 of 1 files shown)


JerryNixon:
Great addition.

PR #656, opened 2016-02-15
Modified the toastservice/helper

abubberman:
1) Refactored the creation of the Toast into separate classes/methods. Also reduced the amount of duplicate code
2) Added the alttext for the image toast
3) Fixed issues in the image and text toasts.

I have tested all toast types. ToastText04 and ToastImageAndText04 seem to ignore the provided content and return a default formatted toast. Not sure why..

All other types are working as expected.

### DIFF
--- a/Template10 (Services)/Template10 (Services).csproj
+++ b/Template10 (Services)/Template10 (Services).csproj
@@ -166,6 +166,7 @@
     <Compile Include="ShareContractService\ShareContractHelper.cs" />
     <Compile Include="ShareContractService\ShareContractService.cs" />
     <Compile Include="ToastService\IToastService.cs" />
+    <Compile Include="ToastService\ToastContent.cs" />
     <Compile Include="ToastService\ToastHelper.cs" />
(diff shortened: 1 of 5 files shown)


JerryNixon:
You don't prefer the method in the toast sample?

abubberman:
Welll... I did not know that there was a nuget package that could replace the ToastService...
After fixing and refactoring these classes I found a MSDN blog where this nuget package is used.
So i've modified my app to use this NuGet package. But that is definitly a better solution than mine :-) 

The Toast classes could be removed I guess?

http://blogs.msdn.com/b/tiles_and_toasts/archive/2015/08/20/introducing-notificationsextensions-for-windows-10.aspx

JerryNixon:
Yeah, probably so. You are like the 10th person to PR them. I will remove them from the services library so people don't keep using them. Wonder how we can get people to see the Samples.

abubberman:
Maybe you can rename the sample tiles to something like TilesAndToasts.  Perhaps this will trigger people to look into it.

Issue #567, opened 2016-01-24
MasterDetail visualState problem


Greywing:
Hi!

Found bug in MasterDetail example:
- run App and select item
  ![selectitem](https://cloud.githubusercontent.com/assets/6736794/12537716/ee682238-c2d6-11e5-91c7-d9a6f8945076.jpg)
- resize App, select another item and resize again
  ![resize](https://cloud.githubusercontent.com/assets/6736794/12537735/7c947a5c-c2d7-11e5-868c-c28295dc5032.jpg)
- try to select another item
  ![selectanothermail](https://cloud.githubusercontent.com/assets/6736794/12537751/d0a53bd6-c2d7-11e5-95b8-9b936373b115.jpg)

JerryNixon:
Looks like Narrow needs to un-select the current item, maybe. Hmm. Not really sure. I will have to look at it. I have a Science Fair on Tuesday and I am writing a Kinect app for the school. What I am saying is that I might not be able to look into this until after Tuesday or so. :) Anyone else, @Windows-XAML/contributors?

JerryNixon:
Fixed.

DanTheManSWE:
This bug is not resolved.
The problem is that when you select an item in ListVisualState, you will go to DetailVisualState, but if you then resize your window, you will still be in DetailVisualState. 

So if you then resize the window again to go to NarrowVisualState, you're still in DetailVisualState, and nothing will happen if you click items in the list until you press the backbutton to go back to ListVisualState. 

I've been trying to find a way to go back to ListVisualState when the adaptivestate triggers, but I've only been coding since Christmas, and my xaml/c# knowledge is limited.

DanTheManSWE:
I think I found a solution, I'll issue a Pull Request soon,

PR #662, opened 2016-02-17
Fix issue #567

DanTheManSWE:
Go back to ListVisualState when MailList SelectedIndex is set to -1.


DanTheManSWE:
I'm not too familiar with GitHub, So if my commits are a bit weird, you know why ;)

PR #665, opened 2016-02-17
Text Box Delay Behavior

Opiumtm:
Delayed actions trigger after a text change in associated text box. Actions is triggered only after a last change. If user type more text before delay time period exhause, delay is renewed.
So, this behavior is very useful in a search text boxes. This behavior let user type text and search can be applied only after user finish typing. This behavior is very useful for an optimization purposes.

I created this behavior for my own project and it work fine.


JerryNixon:
This is a great suggestion, however, the TimeoutAction already exists. Use that with the KeyBehavior and what you want can be composed within the existing library.

Opiumtm:
@JerryNixon No, no! This behavior is quite different from timeout action! 
Just see a code. Maybe my English is bad (because I am from Russia and English is not my native language), but I try to explain you what this behavior essentially do.

This behavior (as I wrote in pull request description) is created for an **optimization** purposes.
It:
- Does not duplicate action calls.
- Does not use DispatcherTimer at all. Just plain `await Task.Delay(...)`

Most important of all, this behavior **does not call actions twice (and more)** when TextBox change events raised. When text in a box change again _before a timeout expire_, TextBoxDelayBehavior _reset delay_.

I already know about TimeoutAction and KeyBehavior. But these classes are too generic and does not cover **specific** use cases. With TimeoutAction contained actions are executed just after a some timeout, so if TextBox raises many TextChanged events (and it do so), **every** TextChanged event will lead to execution of actions after a timeout delay.

Such a behavior (duplicate action execution) is extremely unsatisfactory when cost of delayed action execution is too high (just think about network request to pull filtered information from a remote server...). Even in local scenarios involving local database search (most often SQLite) such behavior is lame.

In this pull request is a right solution to a problem when delayed action itself have a significant execution cost and must not be executed too many times, must be executed only **once** after user is finished typing or take a pause in typing.

So as I stated in description, this pull request is very well suited for an optimization purposes and cover a scenario when action execution cost is too high and it is extremely undesirable to call an action after an every typed charater even after a delay. If delay to an action execution is applied this is not mean that action is not called multiple time.

This behavior **prevents excessive actions execution** which is very critical if action execution cost is high - involves network request, local database search, even a local file reads (it may be fast on a desktop computer, but it is certainly not too fast on a smartphones reading from a SD card)

JerryNixon:
Okay. I see the diff. Let me think about this.

JerryNixon:
Okay. Ready? Here's what we need. I want you to update your PR, please. If you are not interested in coding this new behavior, just let me know and I will add it to my list. Using Reactive Extensions. 

**ThrottledEventTriggerBehavior**

> The Throttled Event Behavior attaches to a target's event, and executes child actions at a rate no faster than the throttle set (in milliseconds). 
- [ ] object ThrottledEventBehavior.Target
- [ ] string ThrottledEventBehavior.Event
- [ ] int ThrottledEventBehavior.Throttle (in milliseconds)
- [ ] int ThrottledEventBehavior.MaximumOccurences 

> RX Sample: http://rxwiki.wikidot.com/101samples#toc30
> 
> RX Docs: https://msdn.microsoft.com/en-us/library/hh229298(v=vs.103).aspx also https://msdn.microsoft.com/en-us/library/hh229400(v=vs.103).aspx
> 
> Event behavior reference: https://github.com/Microsoft/XamlBehaviors/blob/master/src/BehaviorsSDKManaged/Microsoft.Xaml.Interactions/Core/EventTriggerBehavior.cs

``` csharp
//****************************************************************************************************//
//*** Use the throttle operator to ONLY deliver an item that occurs with a 2 second interval       ***//
//*** between it and the next item in the sequence. The throttle buffer will hold an item from the ***//
//*** sequence waiting for the 2 second timespan to pass. If a new item is produced before the     ***//
//*** time span expires, that new item will replace the old item in the buffer and the wait starts ***//
//*** over. If the time span does expire before a new item is produced, then the item in the       ***//
//*** buffer will be observed through any subscriptions on the sequence.                           ***//
//***                                                                                              ***//
//*** To be clear, an item is not guarnteed to be returned every 2 seconds. The use of throttle    ***//
//*** here does guarntee that the subscriber will observe an item no faster than every 2 sec.      ***//
//***                                                                                              ***//
//*** Since a new email is generated at a random time within 3 seconds, the items which are        ***//
//*** generated with 2 seconds of silence following them will also be random.                      ***//
//***                                                                                              ***//
//*** The timers associated with the 2 second time span are run on the .NET thread pool.           ***//
//****************************************************************************************************//
```

**Remember, Eugeny, keep it simple! :)**

Opiumtm:
@JerryNixon I'm not fully understand what you mean :) I thinked about conversion of Behavior into Action with a list of subactions (just like TimeoutAction). Just because it will be more universal (not hardlinked to TextBox and TextChanged event). You mean this? If it is - I will create such Action using same core async method (using `Task.Wait(...)`, timestamp and IsReset flag). Core logic for this behavior is quite simple and elegant, use most basic async language features :) I can convert Behavior into a more appopriate Action.

Opiumtm:
@JerryNixon Well, you mean to create a variation of EventTriggerBehavior with throttling capability?
Anyway I think, to use RX for this is quite excessive - because for now Template10 does not depndent on RX library. Core throttling logic is as simple as it can be, no need to reference RX library for such a simple and clear logic :)

JerryNixon:
Because this logic is responding to something, it is a behavior and not an action. An action is intended to do something. As a result, this is a result responding to the events. When it is ready to do something, it will invoke its child actions. This makes it far more generic and usable.

Yes, I mean to create a variation of EventTriggerBehavior. And, yes, I think using custom logic, at least at this point, is sufficient - so we can skip Reactive Extensions. Though if I were writing this I would use RX almost certainly. You are not required to.

This making sense?

Opiumtm:
@JerryNixon yes, make sense.
As for RX, general purpose library should not be dependent on too many libraries without a significant reason. If external dependency can be avoided - it must be avoided. Because it can create a dependency hell and indirect software bloat (for one small function you pull a thousand nuget packages and any package pull other and there is geometric progression). It should be avoided, and external frameworks should not be used for merely a utilitarian reasons when you create you own framework.

Opiumtm:
@JerryNixon Done improving pull request. Also, updated BehaviorSample application to demonstrate this new behavior at work.

There is little changed copy/paste of `EventTriggerBehavior` named `EventTriggerBehaviorBase` as not sealed class with virtual protected `OnEvent(...)` method so new behaviors can descend from `EventTriggerBehaviorBase` and include their own custom logic on every event.

There is `EventThrottleHelper` class encapsulating throttling logic and it can be used everywhere.

There is `ThrottledEventTriggerBehavior` class, descendant of `EventTriggerBehaviorBase` implements throttling logic via `EventThrottleHelper` and override of virtual protected method `OnEvent(...)` of a base class.

Throttling have two parameters.
`public int Throttle` 
throttle timer in milliseconds, default _1000_

`public bool ResetTimer`
if _true_, throttle timer is reset after each event (usable for TextChanged event of TextBox, so throttled event should be triggered only after user finish typing or take a pause in typing for a throttle period)
if _false_, throttle timer is not reset after each event, so throttled events are fired on a Throttle period schedule when external events are triggered.
default _false_

JerryNixon:
I will review it soon. I am traveling this week.

Opiumtm:
@JerryNixon ping

PR #707, opened 2016-02-29
cleaning up file service

bc3tech:
I was in here to see if there's a problem with re-reads of the same file. I noticed that the signatures for some of the methods were misnomers; offering optional `options` params that aren't used w/in. I cleaned that up, and also "fixed" the naming of method vars to not have a leading `_` as I was confused AF at first when I opened the file and thought we were keeping around a `StorageFile` reference as a private var.


JerryNixon:
Good timing, I plan to introduce this to the main library soon.

PR #738, opened 2016-03-07
MonitorUtils instance is strongly referenced in WeakReference<>.Event…

mitkodi:
WeakReference<MonitorUtils, DisplayInformation, object>.EventAction strongly references MonitorUtils instance.


JerryNixon:
Great fix

Issue #705, opened 2016-02-29
Performance degradation because of NavButtonBehavior and possible memory leaks


Opiumtm:
https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Behaviors/NavButtonBehavior.cs

After investigation with JetBrains DotTrace profiler it is revealed that `NavButtonBehavior` too often executes
`private void Calculate(bool allow = false)` method. And, this method adds a significant performance penalty. Issue show itself in a most terrible manner when Page contains ListView and user is scrolling.
Profiler trace show that it is caused by `private void LayoutUpdated(object sender, object e) { Calculate(true); }` line of code, which call `Calculate` method.

When user scrolls in a ListView, layout is updated too often (ListView is virtualizing items and so items are often recreated and repopulated). It cause major performance penalty to an application that use ListView in its layout. 

Also, NavButtonBehavior attach events to a global singleton Frame, which cause a certain memory leak and further performance degradation after user navigate between pages some time. `Frame.SizeChanged -= SizeChanged;` in a `Detach` method does not work because it is not called automatically on page unload.

`NavButtonBehavior` is used in a `PageHeader` control tempate
https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Themes/PageHeader.xaml
so it attach itself into every page and worse, it buried deep inside a control template and can not be detached by hand without a fragile tricks with visual tree 

In practice it cause a progressive performance degradation after navigations between pages. Scrolling in a ListView suffer a progressively violent lags and this lags get worse after each new page navigation, because already performance inefficient `NavButtonBehavior` does not detach from a global Frame `LayoutChanged` event and this performance cost is multiplied with each new page navigation.
Also it is need to be investigated with memory profiler if `NavButtonBehavior` can cause a global memory leak, preventing pages (which often contain view models and lots of other data inside) from be garbage collected.

Again, it do not much harm to an applications running x86 Desktop (because desktop environment have much more powerful CPU and do not constrained in free memory), but on a resource constrained ARM device it seems to be a disaster.

Opiumtm:
I suggest use of some event throttling technique (or cease use Frame `LayoutUpdate` event) to reduce performance penalty. For global events I suggest use of WeakEvent pattern. It is generally  not wise to attach event handlers to a global singleton objects events and it is a most common cause of memory leaks in a .NET applications. To counter this risks WeakEvent pattern is strongly recommended when attaching to a global singleton object events.

[...]

PR #740, opened 2016-03-08
Update throttling on invocation of Calculate method in NavButtonBehavior

### LINKED ISSUE #725: NavButtonBehavior.cs deadlock in 1.1.7
Latest changes to NavButtonBehavior.cs cause deadlock. I am not sure about the root cause, but application locks up in "NavButtonBehavior.LayoutUpdated", Line 61, while several other threads stopping in CalculateBackVisibility/CalculateForwardVisibility.

Commenting out everything from "NavButtonBehavior.FrameChanged" method fixes this issue, but breaks NavButton, of course.

schuettecarsten:
Remove throttling on invocation of Calculate method, call Calculate method directly. This also makes the reference to RxMain unneccessary. Fixes #725, but might have other side effects, because I do not know why the throttling was introduced.

### DIFF
--- a/Template10 (Library)/Behaviors/NavButtonBehavior.cs
+++ b/Template10 (Library)/Behaviors/NavButtonBehavior.cs
@@ -8,7 +8,6 @@ using Microsoft.Xaml.Interactivity;
 using Template10.Utils;
 using System;
 using Template10.Common;
-using System.Reactive.Linq;
 
 namespace Template10.Behaviors
 {
@@ -35,11 +34,6 @@ namespace Template10.Behaviors
             {
                 _dispatcher = Common.DispatcherWrapper.Current();
 
-                // throttled calculate event
-                var observable = Observable.FromEventPattern(this, nameof(DoCalculate));
-                var throttled = observable.Throttle(TimeSpan.FromMilliseconds(1000));
-                throttled.Subscribe(x => Calculate());
-
                 // handle click
                 element.Click += new Common.WeakReference<NavButtonBehavior, object, RoutedEventArgs>(this)
                 {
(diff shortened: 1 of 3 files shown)


f4lke:
See #705 from 8 days ago.

Opiumtm:
@carstenschuette 
There is more `async/await` friendly class for handle throttling (I write it for similar purposes and use it in throttling event Behavior).

https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Common/EventThrottleHelper.cs

Suggest use it here and remove dependency on RX because RX here is misused.

JerryNixon:
Can we update this @carstenschuette  so it replaces throttling, not removes it? It is still quite important for mobile performance.

schuettecarsten:
Yes. I hope I can do this on the weekend.

schuettecarsten:
@Opiumtm Just took a look into your code. I think isWaiting and isRefreshed in DelayAction should be volatile because they might be accessed from different threads and their value should be shared between threads. Or will "DelayAction" always run from the UI thread?

schuettecarsten:
@Opiumtm , @JerryNixon : Unsure. Works here (on desktop), please review.

Opiumtm:
@carstenschuette it accessed only from UI thread, in the event trigger method there is call dispatching via DispatcherWrapper or check if called from UI thread and fail if not (there are two methods, one dispatches call, other just check if call is made from UI thread). Event throttling helper itself created as DependencyObject with dependency properties, so it can only be accessed from UI thread.

schuettecarsten:
Removed the unneccessary volatile.

PR #757, opened 2016-03-15
Adding Clear() method to IStateItems.

mitkodi:
It could be helpful (e.g., after signing out).


JerryNixon:
Why?

mitkodi:
- To be able to clear all current session state with one call instead of removing items in loop or explicitly casting SessionState to Dictionary. Possible scenario, when the current user signs off.
- It doesn't cost nothing, because SessionState already implements that.

PR #765, opened 2016-03-19
Fixes the momentary flash of RootGrid background brush when switching…

dg2k:
Fixes the momentary flash of RootGrid background brush when switching dark/light themes.

Further info from [Line 703](https://github.com/Windows-XAML/Template10/pull/765/files#diff-ae3950648be7d73717351e18b582c74dR703) in source code as summary/remark documentation.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -182,7 +182,7 @@
         </ResourceDictionary>
     </UserControl.Resources>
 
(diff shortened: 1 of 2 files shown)


JerryNixon:
Is this level of added complexity worth the problem we are solving?

toituu:
I'm not commenting on the complexity, but I second the patch as it does seem to fix quite noticeably flash especially with dark theme...

toituu:
@dg2k might I suggest removing setting the Opacity to 0.95 for the HamburgerButtonGrid. It kind of sticks to your eye if using app in wide mode...

JerryNixon:
Hmm. Okay, still considering this.

PR #774, opened 2016-03-21
Update Newtonsoft.Json to 8.0.3

schuettecarsten:


JerryNixon:
Why?

schuettecarsten:
Because 8.0.3 is the latest version and I prefer to use the latest versions - there have been several minor fixes and will be in future versions. I am always confused when a library references an older version of a library, being unsure if the older version is really needed (like it is on Behavors at the moment) or if I can update to newer versions.

Issue #769, opened 2016-03-20
Follow-up to #766 (Tooltip to NavMenu Item)


dg2k:
@darenm - Further to #766, Tooltip display only when HamburgerMenu is closed can be done (I've tested it already) and HamburgerMenu  will be one better than 1st part apps ;-)

Requires:
- Subscription to HamburgerMenu.PaneOpened and HamburgerMenu.PaneClosed events;
- Dependency property to control tooltip visibility based on the above events;
- Dynamic binding to tooltip string properties;

Since this involves quite a bit, perhaps a good idea as a sample PR which I am happy to do (completed already).

dg2k:
@JerryNixon - You reopened this issue as an enhancement. Can you comment why? The reason I am asking is I thought @o0rebelious0o tooltip implementation is better than what I suggested here.

Issue #766, opened 2016-03-19
Add tooltip to NavMenu Item for hover in collapsed state


darenm:
The News app and others show a tooltip when hovering over a nav menu item. Adding a tooltip to the content in the existing implementation does not show (probably blocked by mouse over states in the control). Needs to be added to HamburgerButtonInfo as a property and added to the template.

dg2k:
@darenm - the property **ToolTipService.ToolTip="tooltip"** can be  readily used with HamburgerButtonInfo (it works fine for me just adding the property). The only caveat is that it should show only when **HamburgerMenu.IsOpen** is **false** (pane closed), else the tooltip would be redundant with the menu text visible. As important, it should, if possible, also show for disabled menu items which by default is suppressed if menu item is disabled. The design is therefore how to temper **ToolTipService.ToolTip="tooltip"**, I think. Saying that there may be a shortcut?

darenm:
Interesting - I must have added it in the wrong place in my haste. Regarding whether it should be shown when the menu is open - right now the 1st part apps show the tooltip regardless of open or closed, but I certainly agree it is redundant in the expanded state if the tooltip displays the same information as the label..

darenm:
Yup - sure enough - adding to the content doesn't work (I was trying to add it just to the icon), but adding it to the HamburgerButtonInfo works like a charm. I'll close this then as fundamentally we're aiming for 1st party app alignment.

PR #698, opened 2016-02-27
Hamburger Menu Custom Styling

vivekhnz:
I've added support for specifying a custom hamburger button style and a custom nav button control template so that developers have greater control over the hamburger menu's visuals.

I've also added brushes for checked+hover and checked+pressed states to the hamburger menu so the hover and pressed colors can be different for the checked state v.s. the unchecked state.

PR #781, opened 2016-03-23
Minimal example of tooltips on Hamburger Menu Items, only when the ha…

rcbevans:
…mburger menu is closed

Based on #766 and #769 I thought it would be cool to have a go at one-upping the first party apps by getting the tooltip showing for Hamburger menu items, only when the Menu pane isn't open.

I'd be happy to see what people think of this approach, refactor and make a real pull request.

Right now, I get the HamburgerButton label by manually checking the children of the Content UIElement in the HamburgerButtonInfo, but adding a property would be nicer so you can specify the tooltip directly, or it defaults to null which hides the tooltip.  Depends how automated people think it should be...  I also thing it would be nice if the "ToolTip" property, if set would set the screenreader text for accessibility purposes.

Thoughts?

_I'm not intending for this to get pulled, but to get some feedback and refactor as appropriate_


rcbevans:
@darenm @JerryNixon @dg2k, you guys have commented on this area previously.  Any thoughts?

dg2k:
@o0rebelious0o @darenm @JerryNixon - my implementation is confined to the Shell.xaml and its codebehind. I don't think it is necessary to change the T10 library code at all. I will post mine as a Sample project PR.

rcbevans:
Well sure, as the  developer you always have the choice to roll your own or make use of inbuilt features.  You could create the whole hamburger menu yourself if you wanted with whatever features you want; but my understanding was T10 is supposed to be a shortcut to provide developers a nice route to get first part app parity on basic features such as title bars and hamburger menus, which to me includes the tool tips and accessibility features

I did it your way first, modifying each hamburgerbuttoninfo to add the tooltip that way, and sure it works, but you have to repeat it for each button (depending how you implemented it exactly), but it kinda defeats the point.  The same goes for accessibility options like narrator, everyone could spin their own, of course they could, but  it's something the library should accommodate IMHO, and with minimal effort and understanding of how the library works underneath required.

dg2k:
@o0rebelious0o @darenm @JerryNixon - Points well made. I agree with what you say in that T10 should absorb features considered essential and yes I also agree that a tooltip is such an essential feature to be built-in. Having checked your proposed implementation, I like it much. I therefore retract my comments.

By way of feedback, is it possible to show the tooltip for a disabled menu item? It would be nice if possible, because the user still wants to know what a disabled menu item does. In my case I  mimicked disabled by graying out a menu item but exiting immediately when menu item is tapped. (Looks disabled but actually not, functionally disabled because tapped event executes conditionally - not neat but works).

Sorry late addition:
- Change "DependencyParameter" to nameof(DependencyParameter) in DPWhenValueConverter.cs
  -  ConverterParameter="Potato" ??

rcbevans:
haha, parameter="Potato".... whoops, I was using the for debugging and forgot to take it out :laughing: 

I'll use `nameof` instead of the magic string, I think that was an autogenerated snippet which I didn't update.

I'll look at how the tooltip behaves when things are disabled; really IMO the buttons shouldn't be using a click event through to codebehind, they should be bound to a delegatecommand which can disable them properly with a predicate.  The tooltip itself is attached to the radiobutton of the template, so I _think_ it should show regardless of the underlying button state, but again, would have to check.

Whilst I'm looking at it, I'll also look at accessibility such as tabstops and narrator, these are really important for any professional quality apps building on T10.

rcbevans:
I have merged the code from #698 into this PR.  I'll do some testing and make sure things are still working okay.

If someone else would like to try and do the same, then that would be great :smile:

rcbevans:
I've done some testing and can see a few issues.  Tooltips on the nav menu items no longer appear to be showing on the latest rs1 insider build.  It appears the binding statement

``` cs
        <converters:DPWhenValueConverter x:Key="DPWhenValueConverter"
                            DependencyParameter="{Binding ToolTipText, Mode=OneWay}"
                            Otherwise="{x:Null}">
```

binding to ToolTipText is no longer working... I can't see the binding reading the property from the HamburgerMenuItem, or being set on the converter.  Whenever the `Convert` method is invoked on the converter, I can see `DependencyParameter` is always null... Setting to a string works fine, so the issue is with the binding :/

Some other observations, tabbing around in accessibility view, if a nav menu entry is highlighted, you can't hit enter to invoke the command on the item, you have to use spacebar; this was the case previously, does anyone have any ideas why?  When a nev menu item is invoked, it seems to get "selected" several times, so the narrator repeats two or three times "selected - home - radio button" which sounds odd.

It's a bit late now, so I'll have to have a look at this again another time; if anyone else has any ideas, that'd be great.

Template10 (Library)/Controls/HamburgerButtonInfo.cs
```diff
@@ -83,6 +79,14 @@ public bool IsEnabled
             DependencyProperty.Register(nameof(IsEnabled), typeof(bool),
                 typeof(HamburgerButtonInfo), new PropertyMetadata(true));
 
+        public string ToolTipText
```
JerryNixon, line 82: First, please note that ToolTip is a ContentControl and not a Text control. This means developers can choose to put XAML controls into it, not just a string. Creating a ToolTipText property instead of using type `object` means you limit developer options. We don't want to do that.

Template10 (Library)/Controls/HamburgerButtonInfo.cs
```diff
@@ -83,6 +79,14 @@ public bool IsEnabled
...
+        {
+            get { return (string)GetValue(ToolTipTextProperty); }
+            set { SetValue(ToolTipTextProperty, value); }
+        }
+        public static readonly DependencyProperty ToolTipTextProperty =
+            DependencyProperty.Register("ToolTipText", typeof(string), typeof(HamburgerButtonInfo), new PropertyMetadata(null));
```
JerryNixon, line 88: This is just a style comment to use nameof(ToolTipText) instead of the literal "ToolTipText" which, of course, would break if the property name changes, though we would not get a compile error. Make sense? Just good practice, is all. I know the snippet did this.

Template10 (Library)/Controls/HamburgerMenu.xaml
```diff
@@ -54,28 +54,45 @@
...
-                             AutomationProperties.Name="Hamburger Menu Item"
                              Command="{Binding NavCommand, ElementName=ThisPage}"
                              CommandParameter="{Binding}" GroupName="Nav" Holding="NavButton_Holding"
                              IsChecked="{Binding IsChecked, Mode=TwoWay}"
                              IsEnabled="{Binding IsEnabled, Mode=TwoWay}" Loaded="NavButton_Loaded"
                              RightTapped="NavButton_RightTapped" Tag="{x:Bind Content}"
                              Tapped="NavButton_Tapped"
-                             ToolTipService.ToolTip="{Binding Path=(ToolTipService.ToolTip)}"
-                             Visibility="{Binding Visibility, Mode=TwoWay}">
+                             ToolTipService.Placement="Top" IsTabStop="True"
```
JerryNixon, line 66: Why Top?

Template10 (Library)/Controls/HamburgerMenu.xaml
```diff
@@ -54,28 +54,45 @@
...
+                             Visibility="{Binding Visibility, Mode=TwoWay}"
+                             AutomationProperties.AccessibilityView="Control"
+                             AutomationProperties.Name="{Binding ToolTipText, Mode=OneWay}"
```
JerryNixon, line 69: Doesn't the screen reader already read the tooltip if the name is empty?

Template10 (Library)/Controls/HamburgerMenu.xaml
```diff
@@ -54,28 +54,45 @@
...
+                                            DependencyParameter="{Binding ToolTipText, Mode=OneWay}"
+                                            Otherwise="{x:Null}">
+                            <converters:DPWhenValueConverter.When>
+                                <x:Boolean>False</x:Boolean>
+                            </converters:DPWhenValueConverter.When>
+                        </converters:DPWhenValueConverter>
+                    </RadioButton.Resources>
+                    <ToolTipService.ToolTip>
+                        <Binding Path="IsOpen" ElementName="ThisPage" 
+                                            Converter="{StaticResource DPWhenValueConverter}"/>
```
JerryNixon, line 82: Yuck. Thank goodness that once we are passing {binding} to a ContentPresenter you will not need this horrible approach. No offense. You did solve it. There zero chance I want to introduce a new converter and zero chance I want to introduce using more converters. The performance implication of a converter is the reason. I would rather introduce an internal, new property to the IHamButtonInfo. But, that's moot now, I hope.

Template10 (Library)/Controls/HamburgerMenu.xaml
```diff
@@ -168,9 +187,10 @@
                                             <Rectangle x:Name="CheckedBack"
                                                        Fill="{Binding NavButtonCheckedBackground, ElementName=ThisPage}"
                                                        IsHitTestVisible="False" Visibility="Collapsed" />
-                                            <ContentControl x:Name="NormalFore" AutomationProperties.AccessibilityView="Raw"
```
JerryNixon, line 171: Good, but please ensure the ToggleButton does not appear in the control list of the narrator now that you have removed this raw value.

Template10 (Library)/Controls/HamburgerMenu.xaml
```diff
@@ -168,9 +187,10 @@
...
+                                            <ContentControl x:Name="NormalFore"
                                                             Content="{TemplateBinding Content}"
                                                             Foreground="{Binding NavButtonForeground, ElementName=ThisPage, FallbackValue=White}"
+                                                            IsTabStop="False"
                                                             IsHitTestVisible="False" />
```
JerryNixon, line 194: Why did you add this? What did this solve?

Template10 (Library)/Controls/HamburgerMenu.xaml
```diff
@@ -181,15 +201,112 @@
                 </RadioButton>
             </DataTemplate>
 
+            <ControlTemplate x:Name="DefaultNavButtonTemplate" TargetType="ToggleButton">
+                <Grid x:Name="RootGrid"
```
JerryNixon, line 205: GitHub is so weird. Of course, you didn't add this whole new template.

Template10 (Library)/Controls/HamburgerMenu.xaml
```diff
@@ -247,12 +364,10 @@
             </Interactivity:Interaction.Behaviors>
         </ContentControl>
 
-        <SplitView x:Name="ShellSplitView" Grid.Column="0" AutomationProperties.Name="Jerry Nixon"
```
JerryNixon, line 250: Wow. Thanks. That was an oversight when I was doing some accessibility testing. It's also, by the way, the root of a bug on the backlog in the SplitView for accessibility. There's an extra Pane that can't be labeled in the tree. But this was supposed to be removed. Thanks. It was invisible though to the readers.

Template10 (Library)/Controls/HamburgerMenu.xaml.cs
```diff
@@ -102,6 +97,12 @@ public HamburgerMenu()
                     // in case the developer has defined zero buttons
                     if (NavButtonCount == 0)
                         _areNavButtonsLoaded = true;
+
+                    // set component styles to default style resources if developer hasn't defined a custom style
+                    if (HamburgerButtonStyle == null)
+                        HamburgerButtonStyle = (Style)Resources["DefaultHamburgerButtonStyle"];
```
JerryNixon, line 103: You want to use Template10.Utils.XamlUtils.GetResource()? Yes.

Template10 (Library)/Converters/DPWhenValueConverter.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Data;
+
+namespace Template10.Converters
+{
+    public class DPWhenValueConverter : DependencyObject, IValueConverter
+    {
```
JerryNixon, line 8: Glad to see this go. That being said, I think we could get away with updating the existing WhenValue converter with propdp instead of CLR props, right? You did this for binding sake? Maybe just inherit from BindableBase?

Template10 (Library)/Utils/ColorUtils.cs
```diff
@@ -10,9 +10,10 @@ public static class ColorUtils
         // this has all been moved to private bacause I am unhappy with the API 
         // and don't want anyone using it (except the HamburgerMenu) until we fix it.
 
-        static Color nearBlack = Colors.Black.Lighten(Accents.Plus20);
-
-        static Color nearWhite = Colors.Black.Lighten(Accents.Plus80);
+        //static Color nearBlack = Colors.Black.Lighten(Accents.Plus20);
+        //static Color nearWhite = Colors.Black.Lighten(Accents.Plus80);
+        static Color nearBlack = Color.FromArgb(255, 51, 51, 51);
+        static Color nearWhite = Color.FromArgb(255, 204, 204, 204);
```
JerryNixon, line 16: Yeah. That's better. Thanks.

Templates (Project)/Hamburger/Views/Shell.xaml
```diff
@@ -12,12 +12,11 @@
       mc:Ignorable="d">
 
     <Controls:HamburgerMenu x:Name="MyHamburgerMenu">
-
         <Controls:HamburgerMenu.PrimaryButtons>
             <!--  mainpage button  -->
-            <Controls:HamburgerButtonInfo ClearHistory="True" PageType="views:MainPage" 
-                                          AutomationProperties.Name="Jerry"
```
JerryNixon, line 19: Oops another accessibility left-over. Thanks for cleaning that up.

Templates (Project)/Hamburger/Views/Shell.xaml
```diff
@@ -29,8 +28,10 @@
 
         <Controls:HamburgerMenu.SecondaryButtons>
             <!--  settingspage button  -->
-            <Controls:HamburgerButtonInfo x:Name="SettingsButton" PageParameter="0"
-                                          PageType="views:SettingsPage">
+            <Controls:HamburgerButtonInfo x:Name="SettingsButton"
+                                          PageParameter="0"
+                                          PageType="views:SettingsPage"
+                                          ToolTipText="Settings">
```
JerryNixon, line 34: Since I am at the end, I am going to comment on one more thing. And that's doing this SAME thing with automation NAME. That's a very important one to carry through and I think it is reasonable that the tooltip would be DIFFERENT than the name in many scenarios. Not the simple ones, of course. Agree? Just another property. Name should be a string, but ToolTip should be object - maybe UIElement.

JerryNixon:
I think it might be easier to close this PR and let you merge in the other PR with @vivekhnz with this one as a new PR altogether. I commented on the code a little, too. Hope it makes sense. You can hit me on Skype at "mr.jerrynixon" if you want to chat about it.

JerryNixon:
Another thing to note, depending on the context of their use, dependency properties without a change event handler do not always raise the clr change events you expect in bindings.

Template10 (Library)/Controls/HamburgerMenu.xaml
rcbevans, line 194: By default it will end up as a tab stop, so as you tab around it'll move down the xaml tree to each level of the nav button, radiobutton -> togglebutton -> etc.

To the user, the nav button is a single thing for them to click and interact with, so there should be a single tabstop for them to stop on and interact with?

The complexity of this Nav Button xaml hierarchy I think is a bit extreme with the different behaviors of the various nested control types interacting.

Template10 (Library)/Controls/HamburgerMenu.xaml
rcbevans, line 69: No it wasn't when I tested, it was reading "Off - Button" for each nav button

Template10 (Library)/Controls/HamburgerMenu.xaml
rcbevans, line 66: Arbitrary, in the first party apps it seemed to be showing tooltips above

Template10 (Library)/Controls/HamburgerButtonInfo.cs
rcbevans, line 88: Yes, I always try and use C#6 wherever possible, this one slipped through the net with the snippet as you said

Template10 (Library)/Controls/HamburgerButtonInfo.cs
rcbevans, line 82: Hadn't realised that, I'll definitely consider that

rcbevans:
Thanks for the extra input @JerryNixon.  I've responded to a couple of minor points, but I think a larger, nicer rethink in necessary to be pull-worthy.  I'll see if I can get in touch with @vivekhnz and rethink this approach.  I might ping you on skype to see if there isn't a better way to think this.

PR #797, opened 2016-03-31
Delete unused/old nupkg to clean up

robertmuehsig:
This NuPkg is not used anywhere and seems to be pretty old - I'm not even sure why it was there in the first place ;)

It's just a cosmetic repo cleanup pr.


JerryNixon:
Wow. That's a classic.

PR #812, opened 2016-04-05
Add ClearPropertyChanged method to reset PropertyChanged event

schuettecarsten:
Add ClearPropertyChanged method to reset PropertyChanged event. We need this method to be able to remove all subscriptions on PropertyChanged event. This helps us to prevent memory leaks.

### DIFF
--- a/Template10 (Library)/Mvvm/BindableBase.cs
+++ b/Template10 (Library)/Mvvm/BindableBase.cs
@@ -13,6 +13,11 @@ namespace Template10.Mvvm
     {
         public event PropertyChangedEventHandler PropertyChanged;
 
+        protected void ClearPropertyChanged()
+        {
+            PropertyChanged = null;
+        }
+
         public virtual bool Set<T>(ref T storage, T value, [CallerMemberName]string propertyName = null)
         {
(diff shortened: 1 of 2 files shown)


JerryNixon:
Describe a use case.

schuettecarsten:
To minimize memory leaks when using x:Bind, it is useful to remove/unsubscribe everything from as many events as possble when Page and attached ViewModel is unloaded. I need the method to be able to remove all subscribers from the PropertyChanged event.

Opiumtm:
@carstenschuette You should not use reflection here. When you compile this with .NET Native (and native compiling is required for store publication) you will get problems with reflection.
 1. Implicit reflection is harmful because it will interfere with rd.xml reflection settings. If programmer choose to disable reflection in native package, he will get runtime errors. If view model is a system class or class from nuget package, reflection also should fail at runtime because by default in .NET Native compiler reflection is only enabled for classes in solution.
 2. Reflection in .NET Native compiled code is slow. Very slow. You should avoid reflection by all costs.

JerryNixon:
When should developers call this method?

JerryNixon:
Could you give me an example?

PR #813, opened 2016-04-05
Disable LoggingService.Enabled by default

schuettecarsten:
Disable LoggingService.Enabled by default - it should be enabled in DEBUG mode by the developers. This reverts commit f0fc1a3 from last week.


JerryNixon:
Why?

PR #816, opened 2016-04-06
Add Command/CommandParameter to HamburgerMenu

schuettecarsten:
Add Command/CommandParameter to HamburgerButtonInfo. The command will be executed when an item in the HamburgerMenu is tapped.


JerryNixon:
Does this conflict with the Click event handled by the HamMenu?

schuettecarsten:
The code is added to the NavButton_Tapped method, so commands are executed when item is tapped. If "Command" is not set, old behaviour is unchanged.

JerryNixon:
Okay, I've had to add this manually as there are many conflicts with the working branch on my machine. But I did. So, don't misunderstand me closing this without merging.

PR #819, opened 2016-04-07
Feature/fix hamburger menu adaptive triggers in fullscreen

### LINKED ISSUE #811: Hamburger Menu starts with wrong "breakpoint"
When HamburgerMenu is set to "FullScreen=true" in XAML, adaptive triggers for normal/wide/narrow state are not applied when then HamburgerMenu is shown for the first time because UpdateFullScreen() removed the splitview from visual tree.

schuettecarsten:
This PR will fix #811.
It works here, but please review/test before merging.


JerryNixon:
Beautiful work. Why are you skeptical of it?

PR #827, opened 2016-04-12
Update Hamburger Menu controls to support swipe to show on Narrow Status

lyling:


JerryNixon:
Is this really a good idea? won't it get in the way of controls mistakenly placed UNDER this rectangle?

JerryNixon:
Did you remember that the menu could be on the Right?

lyling:
Yes. It may happen.  i will implement it using customize splitview to do it.

any suggestion from you?

JerryNixon:
Just not convinced this is a good idea. Is it?

darenm:
Are there risks that it could swallow interaction activities destined for the content area? Drag and drop, swipe to change pivot, etc.?

lyling:
please reject this pull request , I will implement it by other solution , customize splitview .

PR #843, opened 2016-04-17
initial keyboard navigation for HamburgerMenu

### LINKED ISSUE #806: hamburger menu keyboard control doesn't work
Navigation using the keyboard does not go well for the hamburger menu.  Even if you focus over to it using tab stops up and down do not move between the items.  This is a minor annoyance on desktop but for a keyboard only interface (Xbox controller)  its not very usable.

mitchcapper:
closes Windows-XAML/Template10#806

this is a solution to make keyboard controls more usable in the hamburger menu.  It largely does so through changing the container to a GridView rather than itemview where selection controls are more natural.   Its not perfect as we have to pass the selection down to the child button when the selected item changes.  When you only have one item this is problematic as its not transferred directly (so double tab stops occurs on things like settings when the item count is only 1).  The benefit is it is far more keyboard friendly now.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -242,7 +242,7 @@
 
         <ContentControl x:Name="Header"
                         Height="48"
-                        Margin="48,0,0,0"
+                        Margin="48,0,0,0"  IsTabStop="False"
                         VerticalAlignment="Top"
                         HorizontalContentAlignment="Stretch"
                         Content="{Binding HeaderContent, ElementName=ThisPage}">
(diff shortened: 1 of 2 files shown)


JerryNixon:
I recognize the problem you want to solve. This, at least right now, is not the right solution. The solution we need must support Toggle, Commands, and Literal buttons. A GridView can't do that. Not without considerable customization not part of your PR. Perhaps there is a smaller solution wen can add at this point.

JerryNixon:
Leave a comment

mitchcapper:
Sure, I am happy to hear some feedback it seemed like a good way to go as you are selecting a specific item.  It doesn't have the highlight color like listbox and allows for the buttons to come through.   Its fairly important to figure out a good way to reconcile this.  We could go back to looking at the xbox XYFocus options but they would only work inside the control (which wouldn't be the end of the world).

mitchcapper:
Sorry, and as for why its important right now any app using the menu is completely unusable on a an xbox/keyboard environment  (tab stop could help partially but still not solve directional issue).

mitchcapper:
Making the class unsealed currently would atleast allow for 3rd party changes, any reason to mark the class sealed?

JerryNixon:
I understand. But, we have to support Literal and Command.

mitchcapper:
Sorry to confirm while I change from ItemsControl ->GridView the actual controls implementing the buttons didn't change (still togglebutton) so the command binding still works, I am not sure what you mean by literals.  We can move the conversation back to the bug report rather than PR to come up with another solution.

PR #849, opened 2016-04-19
Minor focus improvements, improves on 64c419a

### LINKED ISSUE #806: hamburger menu keyboard control doesn't work

mitchcapper:
should close Windows-XAML/Template10#806
more details in #806


JerryNixon:
Very nice. I've already made these changes. But brilliant work; keep it up.

PR #853, opened 2016-04-20
Improve serialization of navigation parameters

schuettecarsten:
Improve serialization of navigation parameters by also saving the serialized string and compare the string instead of serialize the saved parameter again. This might cause minor side effects when complex objects used as parameter are modified later, but IMHO complex parameter objects should only used to transfer parameters and never be modified.


JerryNixon:
I agree that this needs to be solved. I will use this PR as reference as we fix the problem. Can't you solve this, in the meanwhile, simply by using SessionState or a custom Repository?

PR #873, opened 2016-04-23
HamburgerButton is now also hidden in FullScreen mode

schuettecarsten:


JerryNixon:
Can't use Visibility since it is already bound. Fixed already.

PR #874, opened 2016-04-24
Added Feedback Hub integration helper

callummoffat:
I thought this would make a nice addition to Template 10. The new Monetization SDK adds support for the new feedback for developers system, which is absolutely wicked. So, since Template 10's job is to make the developer's life easier, I thought adding a helper utility for the Feedback Hub integration would be awesome. Not many additions, but here's the major stuff:
- Added the Store Engagement SDK to the library
- Added a Feedback utility to the Nuget library, which handles the integration.

### DIFF
--- a/Template10 (Library)/Template10 (Library).csproj
+++ b/Template10 (Library)/Template10 (Library).csproj
@@ -199,6 +199,7 @@
     <Compile Include="Utils\ColorUtils.cs" />
     <Compile Include="Utils\DeviceUtils.cs" />
     <Compile Include="Utils\ExpressionUtils.cs" />
+    <Compile Include="Utils\FeedbackUtils.cs" />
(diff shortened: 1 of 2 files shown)


mitchcapper:
Is there a lot of benefit to offering this helper? While the store engagement SDK doesn't really have a lot of downside to including, I am not sure the helper does a lot for users either.

dg2k:
Don't mean to discourage but there are straightforward approaches to this, see for instance [a sample here](https://msdn.microsoft.com/windows/uwp/monetize/launch-feedback-hub-from-your-app), rather than bloat T10 with utilities that devs choose to include.

JerryNixon:
This is a nice addition that will benefit only a few developers. For the rest, this seems unnecessary, and should I be wrong, adding it to an app is pretty easy-peasy. I vote, no. This time.

PR #877, opened 2016-04-24
Custom hamburger menu styles

vivekhnz:
I've added support for overriding:
- `Style` of the hamburger button (added new dependency property called `HamburgerButtonStyle`)
- `Template` of the internal `ToggleButton` within `NavToggleButtonTemplate` (added new dependency property called `NavToggleButtonControlTemplate`)
- `Template` of the internal `Button` within `NavCommandButtonTemplate` (added new dependency property called `NavCommandButtonControlTemplate`)

This allows developers to completely customize the look of navigation menu items and the hamburger button.


JerryNixon:
Possible down-side. Instead of pull requests to make it right, developers will simply implement their own solutions locally. :( Okay. update your submission so I can merge and test it.

vivekhnz:
I've updated my branch so you can merge it in.

I understand why allowing complete customization could be a downside. In my case, I wanted to adjust the specific `Opacity` of the pressed and hover `VisualStates` on the hamburger button and menu items. We could add these as additional properties to the `HamburgerMenu` instead, but I think it might become too bloated.

Since this only allows overriding of the `Styles` and `ControlTemplates`, I think other devs will still submit PRs to fix bugs and add features. This just allows them to fine tune the control to fit their app's branding a little better.

JerryNixon:
Now that there are Foreground & Background for EVERY SINGLE state, does this potentially make the need to do this moot? It gives you control over everything, and with the introduction of Literal, there's even more control over the thing if you want it. Back to the downside of good developers no longer contributing but just locally customizing is a detriment to the framework. What do you think?

vivekhnz:
I see your point. I'll try to explain the benefits I think my PR provides a little better.

The existing `HamburgerButtonStyle` allows developers to customize the `Foreground` and `Background` of the `Button`. The `PointerOver` and `Pressed` states apply a 20% or 40% opacity white overlay respectively over the base `Background` color. There's no way for developers to modify the `Opacity` or the color of either overlay. In the case of my app, the `Background` of the hamburger button is a dark color so the overlays don't appear as subtle as I'd like them to be.

Regarding the menu items, you're definitely correct since devs are able to specify the appropriate `Colors` for every `VisualState`. For my app, I wanted the appearance of the nav menu items to be consistent with the hamburger button. That includes the colors (overlays) for each individual `VisualState`. Implementing the menu item `VisualStates` in the same way as the `HamburgerButton` (using semi-transparent overlays instead of solid `Colors`) ensures a consistent look between the two controls.

Furthermore, my app will allow users to select different app themes. In the current implementation, I'd have to define the `Background` and `Foreground` `Colors` for every single `VisualState`, for every single theme. It's much easier to instead use semi-transparent overlays since I only have to define the underlying 'base' `Color`.

Technically by using semi-transparent overlays, I'm sacrificing customizability (being able to define `Colors` for each individual `VisualState`) for the ability to easily add and modify themes. I recognize that other devs may want the full level of customizability instead however, which is why I think adding support for overriding the default `Style` and `ControlTemplates` is a good way of enabling developers to accomplish this.

There could be other uses as well for overriding the `Styles` in the future as well. Customizing `VisualTransitions` between each `VisualState` isn't possible in the current implementation. Some devs may want to integrate some of the fancy new Windows.UI.Composition effects into their `HamburgerMenu` as well.

JerryNixon:
You're not wrong. Let me think on it.

JerryNixon:
You've done very good work. I hope you can understand why we are not going to go this direction right now. With the introduction of ButtonType=Literal, I hope most of your use cases can be met.

JerryNixon:
OK, after some reconsideration this is pretty much a great idea. And we need to do it, but will wait until version two when we have redstone supported.

vivekhnz:
Alright, sounds good. Would you like me to continue to update the associated branch as we go or wait until after v2?

JerryNixon:
Yes. Let's up date this with the current branch and let's also include a very simple Sample project to demonstrate how to use these new properties and how easy it will be. Make sense?

PR #884, opened 2016-04-26
Public Property wrapper for readonly LoadedNavButtons

dg2k:
```
readonly List<InfoElement> LoadedNavButtons = new List<InfoElement>();
```

in HamburgerMenu.xaml.cs provides wealth of info that is helpful to developers, particularly for implementing runtime theme setting. For instance, it avoids an alternative approach of using VisualTreeHelper to get to ToggleButtons. In short, it contains well-cooked and ready info to delight every developer. It's readonly so ... and this PR just provides a public property to **get** it (... pun is intended ;-)


JerryNixon:
I think this will need to be re-cast as a IReadOnlyList if we are going to do something like this. I am wary of this because developers do not have a State or Event that indicates the List would be ready to read (not changing). Making it public is really not enough. Give me a use case where you would want this.

mitchcapper:
Changing to an observablecollection and inotifypropertychanged but I am not sure of the advantage.  Exposing the FrameworkElement that the HamburgerButtonInfo (could be a property on it) might get @dg2k much the same result (as the user can already reference these things).  I am not sure of how one would directly use these either though as there are fairly deep databindings to do most actions.

schuettecarsten:
I would prefer a method like "GetFrameworkElementForHamburgerButton(HamburgerButtonInfo info)" because then you do not know anything about how HamburgerMenu works internally.

JerryNixon:
I think we might start to consider the Entity Framework approach of GetInfrastructure<T>() and SetInfrastructure<T>() for injection, meanwhile, a list like this is too variable, I'm not sure how to make it reliable for a developer who is calling it. Even a GetFrameworkElement() method could result in null if called too early in the pipeline, right? Needs to be safe.

dg2k:
@JerryNixon - I was trying to post a Sample project which I hope would provide a background to the above use case. I think something I couldn't figure out happened. Bottom-line - never did a PR while another is pending, hence this confusion on my part. I think it's better to post a separate PR for the Sample project (entitled DynamicThemes).

Can you get to the Sample project please?

My comments on the above after the PR.

dg2k:
I fully understand the ramifications in using the proposed Property as it is. Although I am not pressing for it to be available (alternatives exist) it is possible to safeguard its use, for instance as follows:
The following is readily available to the dev.

`var menuCount = HamburgerMenu.PrimaryButtons.Count + HamburgerMenu.SecondaryButtons.Count;`

Then make sure that:
`HamburgerMenu.LoadedNavigationButtons.Count >= menuCount`
for it to be viable.
I must admit I haven't given much thought about the possible implications in the context of other devs (such as the need to ensuring that all NavButtons are fully loaded), but I get it.

I hope the Sample project I am trying to post will explain the interest behind this.

JerryNixon:
The best application might be to implement an AfterNavButtonsLoaded event with AllNavButtons as a read-only argument property. This doesn't, however, account for developers who dynamically add buttons to the Primary and Secondary button collections. And, in a way, leads to misleading results in those cases. Should the event raise again? What about when a button is removed? There are a lot of scenarios to consider here. I think we should wait until it comes to a head and is required.

mitchcapper:
An observable collection could fix this but I think the Functional call is a bit better to abstract things out and an event for NavButtonsChanged or similar.

PR #886, opened 2016-04-27
Make properties virtual in ViewModelBase

schuettecarsten:
Make NavigationService and Dispatcher properties virtual to allow to override getter (or setter) in descendant classes. We need this in our implementations :-)


JerryNixon:
Then do it to all of them.

schuettecarsten:
You're right. Done.

PR #887, opened 2016-04-27
Force visual state for previous toggle button when selection is changed

schuettecarsten:
Sometimes (mostly after the second change), ToggleButton in hamburger menu stay at the wrong visual state and are still shown as checked. As a workaround, with this commit the visual state for the previous toggle button is set to "Normal" when the selection is changed. This fixes the problem here

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
@@ -294,6 +294,15 @@ namespace Template10.Controls
             {
                 previous.IsChecked = false;
                 previous.RaiseUnselected();
+
+                // Workaround for visual state of ToggleButton not reset correctly
+                var fwe = LoadedNavButtons.Where(x => x.HamburgerButtonInfo == previous)
+                    .Select(x => x.FrameworkElement)
+                    .FirstOrDefault() as Control;
+                if (fwe != null)
+                {
+                    VisualStateManager.GoToState(fwe, "Normal", true);
+                }
             }
 
             // navigate only when all navigation buttons have been loaded


JerryNixon:
When does this occur?

schuettecarsten:
Our Hamburger Menu has 7 items, all Toggle and with a target page. The first is default. Klick on second item, first item is set to "Normal". Click on third button, now second button stays on "Pressed" state and is still highlighted. Navigation works as expected.

JerryNixon:
Seems like there might be something else going on here, Carsten. Are you sure this workaround is necessary and there's not something fundamentally out of order otherwise?

schuettecarsten:
It's only a workaround. I do not understand what's going on and why the visual states are applied, I did not find anything in code that sets visual states for Hamburger Menu buttons. But the workaround makes it work here.

dg2k:
@JerryNixon, @carstenschuette - please also see the fixes I've made in #892 which addresses the same issue and more. My suspicion is that there is a bug in ToggleButton's ControlTemplate visual states. This problem reveals itself if you try to set a theme during run-time otherwise you'd rarely notice this.

The fundamental problem is that the **UnChecked** visual state never returns to the Normal (initial) visual state. I introduced a DP (and dummy visual state for command button) for controlling the state transition to work as it should. Curiously, ToggleButton has no explicit **UnChecked** state -- could this be the root of the problem (hence _platform bug_)?

I would appreciate your feedback on the Sample project as I have spent quite sometime on it and provides more insight into some issues that crop up while setting themes during run-time.

**Note:** I made a late typo correction; I previously said Checked when it should be **UnChecked** as shown in bold above.

PR #891, opened 2016-04-27
Update HamburgerButtonGrid 

jp-weber:
So that ContentControl Header can be displayed correctly.

If the Background of HamburgerButtonGrid not transparent, the ContentControl Header is covered.


JerryNixon:
No longer valid

PR #892, opened 2016-04-27
DynamicThemes Sample Project

dg2k:
This is a sample project demonstrating how themes (defined in Styles\Custom.xaml) can be set during runtime. There are a number of fixes in HamburgerMenu control (xaml and codebehind) to enable dynamic setting of themes.


JerryNixon:
Too large of a PR man.

dg2k:
@JerryNixon - It's a sample project. How small should a sample PR be to be considered and not rejected outright? This is not out of line compared with the other samples. Have you looked at the  sample in operation?

PR #897, opened 2016-04-29
PR fixes ControlTemplate visual states to avoid visual anomalies when…

dg2k:
PR fixes ControlTemplate visual states to avoid visual anomalies when setting themes at runtime.

For background:
please see comments in PR #892 and #887. This is a PR confined to the HamburgerMenu control part.
To demo the fixes proposed here, the Sample project in #892 can be used.
If required I can post the PR for the Sample alone.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
@@ -453,5 +453,24 @@ namespace Template10.Controls
                 item.Value.Invoke(e);
             }
         }
+
+        public bool IsSettingToNormalVisualState
+        {
+            get { return (bool)GetValue(SetToNormalVisualStateProperty); }
+            set { SetValue(SetToNormalVisualStateProperty, value); }
+        }
+        public static readonly DependencyProperty SetToNormalVisualStateProperty =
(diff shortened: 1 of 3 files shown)


JerryNixon:
What are the anomalies?

dg2k:
Upon changing the theme, the hover background is left behind when a menu button is unchecked. I think this is what @carstenschuette is also referring to in #887. Also, when changing the theme, the menu button background doesn't get updated.

GIF video is coming up to show the visual anomalies referred to.

dg2k:
**With Visual Anomalies**
Note that when the theme changes, the NavButton background is not updated because the ControlTemplate visual state doesn't reset to initial state. The hover background being left unchanged is not shown here (I have to uncomment code from different places, and thought this would suffice to convince. Video with the visual state fix will follow.  Tap or click the animated GIF for larger screen view.

without-fix

dg2k:
**With Visual States Fixed:**
@JerryNixon - Note that all the color scheme now updated as expected.
Tap or click the animated GIF for larger screen view.

with-fix

schuettecarsten:
Any news on this? What do you think, @JerryNixon ?

JerryNixon:
This seems like a reasonable approach, now that I have reviewed it.

Issue #894, opened 2016-04-28
Snippets not working


alexdresko:
Okay, I'm finally tired of copy/pasting my existing code and the T10 snippets aren't working. 

Proof I've got the extension installed... Is there anything else I need to do? How can I validate the snippets are installed or manually install them? 

![image](https://cloud.githubusercontent.com/assets/333022/14897959/5bd73f50-0d52-11e6-8380-403c8b842504.png)

alexdresko:
Ah, I found the snippets in the repo, as well as InstallSnippets.cmd.. running that manually installed the snippets. Is that the correct process? I'll be happy to update the wiki with instructions if so.

JerryNixon:
Thank you for offering, please do. It would be a help.

PR #899, opened 2016-04-29
Install snippets with VSIX

alexdresko:
After getting approval to modify the wiki to explain how to install the snippets (https://github.com/Windows-XAML/Template10/issues/894), I began to wonder how easy it would be to install the snippets automatically with the VSIX. 

Mads, the VSIX master, explains perfectly how to make it happen in http://madskristensen.net/post/shipping-visual-studio-snippets-in-an-extension. 

So I followed his instructions and it worked perfectly. 

It did require moving the snippets from their current location, but I modified `Snippets/InstallSnippets.cmd` so it continues to work as expected, therefore continuing to allow for manual snippet installation. 

I'll still do the wiki, but wanted to get this in first, since that would also impact what I put in the wiki.


alexdresko:
@JerryNixon For the snippets page on the wiki, when do you plan on releasing an updated installer?

Issue #902, opened 2016-05-02
Performance issue: EllipsisBehavior


Opiumtm:
![image](https://cloud.githubusercontent.com/assets/16596090/14946328/a5abdac8-1031-11e6-9d41-c0e8d036cb1e.png)

Here screenshot from performance profiler.
After some navigation from pages EllipsisBehavior start to lag application especially on mobile devices.

It should be improved.

Opiumtm:
Again, plain hook on `LayoutUpdated` event is wrong. It become a disaster when page layout is changed frequently. It's a case when there is a `ListView` on page. `ListView` is **virtualizing** items control, so it create and destroy list items on demand. Layout is changed any time list is scrolled. If there is multi-phased rendering in `ListView` items for scroll optimization - things are going even worse. Layout updated not only on item rendering but also on every rendering phase.

I think it's **fundamentally wrong** to hook plainly on `LayoutUpdated` event in T10 library.

Opiumtm:
Update: `LayoutUpdated` is a `RoutedEvent`, so we have to check if event source is right control.
Routed events can be catched from other sources and in a case of `LayoutUpdated`, event source should be validated.

Opiumtm:
Here is why hook on `LayoutUpdated` and do some expensive work (like iterating over visual tree to find "More" button) is fundamentally wrong.

https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.frameworkelement.layoutupdated

> When you handle LayoutUpdated, do not rely on the sender value. For LayoutUpdated, sender is always null, regardless of where the handler is attached. This is to prevent handlers from assigning any meaning to sender, such as implying that it was that specific element that fired the event out of the visual tree. **LayoutUpdated implies that something in the overall visual tree has changed**, and each specific object anywhere in the tree has the option of handling this occurrence. If you're familiar with lower-level render API design, you can equate LayoutUpdated being fired as similar to a "redraw needed" flag being set as part of an object-driven, retained-mode rendering logic.
> **Because LayoutUpdated fires in many circumstances and isn't always specific to an object that actually changes**, consider whether handling the SizeChanged event instead is more appropriate for your scenario.

Opiumtm:
@JerryNixon look at it, it seems like a technical flaw in T10 library design
To prevent poor performance another approach needed in controls and behaviors. `NavButtonBehavior` had similar issue and also because of `LayoutUpdated` event

Opiumtm:
[...]

PR #906, opened 2016-05-03
Performance optimization for #902 issue

Opiumtm:
Updated method of `MoreButton`  extraction from templated visual tree.

Replaced **very** expensive call `XamlUtils.AllChildren<Control>(commandBar)` with much optimized methods.

If `EllipsisBehavior` is applied to `PageHeader` control, used cached internal value (obtained and cached in overrided `OnApplyTemplate` protected method of `PageHeader` control). It is the most optimized method because WinRT call is performed only once when control template is applied.

If `EllipsisBehavior` is applied to `CommandBar`, used just 3 WinRT calls instead of very unoptimized manual visual tree walk.

Simple `FindName` works if called on templated visual tree root. No need for expensive manual tree walk with many WinRT interop calls.

```
if (VisualTreeHelper.GetChildrenCount(commandBar) > 0)  
{  
    var child = VisualTreeHelper.GetChild(commandBar, 0) as FrameworkElement; /* Templated root */  
    return child?.FindName("MoreButton") as Button;  
}  
```

### DIFF
--- a/Template10 (Library)/Behaviors/EllipsisBehavior.cs
+++ b/Template10 (Library)/Behaviors/EllipsisBehavior.cs
@@ -3,7 +3,10 @@ using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Template10.Utils;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using Windows.Foundation.Collections;
+using Windows.UI.Xaml.Media;
+using Template10.Controls;
 
 namespace Template10.Behaviors
 {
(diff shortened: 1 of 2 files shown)


Opiumtm:
Tested this fix with real-life application on actual Lumia 950 device. UI performance radically improved. All visual lags and overall poor UI performance goes away.

callummoffat:
My question would be: does this extensive optimization have any impact on stability? Speed shouldn't come at the expense of stability. I love the improved performance, but I'm a little worried about the potential cost to stability.

Opiumtm:
@callummoffat Testing stability on actual device right now, not experienced crashes or incorrect work. Already tested in debugger. Anyway this fix is very conservative in nature. It doesn't change any high-level logic. Just technically improve performance bottleneck found by app profiling. If improved methods fail (they should not fail), fallback original logic is used. Leaved original method as fallback because of conservative approach in this fix.

callummoffat:
That sounds... really good. I'm not seeing any problems with including it (from the perspective of a developer who knows little about T10's underlying code). Maximum performance for minimum risk is a good thing, no matter what way you look at it.

Opiumtm:
@callummoffat Already wasted huge amount of time (couple of months to be honest) optimizing poor app performance to finally find this performance bottleneck. It doesn't critical on x86 desktop and somewhat tolerable on top tier Lumia 950. But on less powerful devices it performed very poor. It takes such a long time to investigate, exclude other possible factors, prepare correct testing conditions (not too easy on less affected x86 Intel desktop) and finally catch root cause in performance profiler, then investigate library code and find fundamentally unoptimized method called from a low-level system layout updating cycle on every visual tree update. It was a crazy adventure in coding and testing :smile:

callummoffat:
@JerryNixon,
This should definitely be included. Huge speed improvements with little or no impact on stability. It would be a shame not to include this.

sebfrie:
I just wanted to second that. Please pull asap.

@Opiumtm Thanks for the adventure. :)

JerryNixon:
Nice work.

PR #940, opened 2016-05-10
Fixes NullReferenceException that is thrown in Full Screen mode startup.

dg2k:
In Full Screen mode, `FullScreenPropertyChanged` fires before HamburgerBackground brush is initialized, and fails with `System.NullReferenceException` at [L432 ](https://github.com/Windows-XAML/Template10/blob/master/Template10%20%28Library%29/Controls/HamburgerMenu.xaml.cs#L432)of HamburgerMenu.xaml.cs. This PR sets the null initial values to default brush values for the Dark Theme.

**Note** the non-null value is non persistent -- only to avoid NullReferenceException and doesn't matter what value;the usual theme setting follows for persisting brush values.

Perhaps a good idea to set all the brushes to their default non-null values? For now, it's only `FullScreenPropertyChanged` that needs fixing.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
@@ -64,7 +64,7 @@ namespace Template10.Controls
         }
         public static readonly DependencyProperty HamburgerForegroundProperty =
               DependencyProperty.Register(nameof(HamburgerForeground), typeof(Brush),
-                  typeof(HamburgerMenu), new PropertyMetadata(null, (d, e) => Changed(nameof(HamburgerForeground), e)));
+                  typeof(HamburgerMenu), new PropertyMetadata(Colors.White.ToSolidColorBrush(), (d, e) => Changed(nameof(HamburgerForeground), e)));
 
         public Brush HamburgerBackground
         {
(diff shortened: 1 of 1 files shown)


mitkodi:
Or set **HamburgerMenu.AccentColor's** default to something (e,g. Application.Current.Resources["SystemAccentColor"]) and call RefreshStyles in HamburgerMenu's ctor.

dg2k:
It is not important what value you set it to as long as it's non-null as this value is not persistent.
The [dev] theme setting comes just after FullScreenPropertyChanged-> UpdateFullScreen() is done.

... a bit intrigued by the above suggestion though -- you're sending back dev XAML resources into HamburgerMenu code ?

mitkodi:
SystemAccentColor is not usually defined by a developer. And yes, I'm not sure at all that it's good (that's why I didn't submit PR).

I think that if you define HamburgerBackground (or better AccentColor) in App.xml (or in a merged resource) or locally at HamburgerMenu you should not get the error.

dg2k:
I have defined HamburgerBackground (which you'd do under Loaded event) but FullScreenPropertyChanged comes too soon. It's a matter of timing not for lack of definition. For instance, if FullScreenPropertyChanged  is delayed it will work. But why bother with timing at all when the culprit null can be avoided?

JerryNixon:
There are more commits around the feature of full screen, then I think anything else in this entire library. Hopefully this will put this feature to bed.

mitkodi:
I think that setting default value for **AccentColor** is better than doing it for **HamburgerBackground/Foreground** etc., because these colors are set in **RefreshStyles** (depending on AccentColor). In such way we'll have a good looking menu out of the box, without applying any local style to the app's menu (which is the case with BottomAppBar sample), instead of menu with visible only a button.

And about the null reference exception, it's enough to check if **HamburgerBackground** is null before setting its Opacity (in **UpdateFullScreen**):

```
if (HamburgerBackground != null) HamburgerBackground.Opacity = opacity;
```

PR #941, opened 2016-05-10
ViewService enhancement to control Lifetime of secondary windows

Viachaslau-Zinkevich:
**Perhaps should wait before stable version is released.**

ViewService enhancement to create secondary views and control their Lifetime. Should be used instead of INavigationService.OpenAsync.
Before using should be statically initialized in App constructor through
`
public App()
 {
    InitializeComponent();
    ViewService.Initialize(this);
}`

Added example to Samples/MultipleViews.


Template10 (Library)/Services/NavigationService/INavigationService.cs
```diff
@@ -23,6 +23,7 @@ public interface INavigationService
 
         void Refresh();
 
+        [Obsolete("Use Template10.Services.ViewService.IViewService.OpenAsync instead")]
```
JerryNixon: Let's consider a way to leave this alone. Could we change the signature to return an instance of IViewService or something to encapsulate the use of the service more simply for developers using T10?

Viachaslau-Zinkevich:
Hm, makes sense. Would you prefer using IViewService inside NavigationService and changed NavigationService.OpenAsync signature to return ViewLifeController?
Also there should be initialization to subscribe for OnWindowCreated. Currently I left this to user to call in constructor, but I would have it done somewhere automatically. Any suggestions?

JerryNixon:
The goal is to keep the current behavior without requiring a developer to learn a new approach, just give the more. We'll need a ViewLife interface.

JerryNixon:
I've change my mind about this I think will go ahead and do it but wait until version 2 when we have redstone supported.

JerryNixon:
I want INavigationService to be the same, for sure. That's used all over the place and in thousands of apps. We can't change that. But we can change the signature of OpenAsync()

https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Services/NavigationService/NavigationService.cs#L186

Since this is only Task, changing the return signature is not a problem. Right? No change to the developer using it, but developers who understand ViewService can get extra capabilities.

JerryNixon:
You okay with that change?

Viachaslau-Zinkevich:
@JerryNixon absolutely. I can utilize ViewService inside NavigationService as delegating call. Would it be okay for you if I add ViewService initialization(the part that subscribes to OnWindowCreated event) somewhere inside, so that users won't have to bother about this thing?
I swear I got a message about a comment where you were talking about Redstone and v2..but don't see it here

JerryNixon:
The more "taken care of" the better. If you can encapsulate all of the work and the developer just gets benefit for free, then we should definitely do it. 

The Redstone message was actually for another thread and I posted it here by mistake. I think I said I would wait until RS1, but that is NOT what I want to do with this. If you get the PR in before Friday, then I would like to include it in the next NuGet release.

Viachaslau-Zinkevich:
@JerryNixon I updated NavigationService to utilize ViewService inside and added some initialization into BootStrapper

JerryNixon:
Very nice. Needs an update to merge, unfortunately.

Viachaslau-Zinkevich:
Updated, hope to see it without conflicts now

Viachaslau-Zinkevich:
@JerryNixon Guess,can be merged now

Viachaslau-Zinkevich:
@JerryNixon would you please notify here, when you will have plans to merge so that I resolved conflicts then?

JerryNixon:
We're in the process of marking all the services as Sealed. Do that, please, and go ahead and sync. The code looks good. Send me an email jnixon@microsoft when you are ready so I can jump on it and you don't have to sync up again.

Viachaslau-Zinkevich:
Hi @JerryNixon , Sorry for delay, I've updated service with sealed modifier and merged

PR #942, opened 2016-05-10
Search sample: fix object reference not set to an instance of an object on startup

mitkodi:
Maybe a typo.


JerryNixon:
Feels **good**  to submit a PR! :) Huh?

mitkodi:
Btw, **BottomAppBar** sample is also broken because **HamburgerBackground** is not set. Maybe different default value or/and a check in **HamburgerMenu.UpdateFullScreen**.

dg2k:
That's what PR #940 is for.

PR #952, opened 2016-05-12
Adds DisplayModeChanged event to HamburgerMenu control.

dg2k:
Okay, as a follow-up to PR #950, we can now add a `DisplayModeChanged` event to the HamburgerMenu control.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.PublicMethods.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.PublicMethods.xaml.cs
@@ -29,6 +29,7 @@ namespace Template10.Controls
         public event EventHandler PaneOpened;
         public event EventHandler PaneClosed;
         public event EventHandler<ChangedEventArgs<HamburgerButtonInfo>> SelectedChanged;
+        public event EventHandler<ChangedEventArgs<SplitViewDisplayMode>> DisplayModeChanged;
 
         public void RefreshStyles(ElementTheme theme, bool clearExisting = false)
         {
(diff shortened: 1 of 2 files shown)


JerryNixon:
DisplayMode is already a dependency property, a developer can already HamburgerMenu.RegisterDepdencyPropertyChanged(DisplayMode). Do you think an event is really necessary?

dg2k:
It makes it neat as an event, just an opinion. The same comment also applies to Selected and SelectedChanged which are already available to devs.

But the more I look into it, there are things to tidy up if the event is to be implemented, or even to use with the RegisterDependencyPropertyChanged.

In fact, as the DisplayMode is managed by VisualState,  for this we need to bind ShellSplitView's DisplayMode **TwoWay** as follows:

`DisplayMode="{x:Bind DisplayMode, Mode=TwoWay}"` 

I have a PR already made if you accept it.

Here is a question for you: In the comment for DisplayMode DP you write:

> This is generally automatically set by the visual states, but can manually be set to accommodate custom layouts.

How is the manual DisplayMode setting accommodated when visual state is in full control, i.e., no option to override visual state's control?

JerryNixon:
Here's where the Visual State otherwise handles the DisplayMode for the developer:
https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Controls/HamburgerMenu.xaml#L285
If you set a Visual State MinWidth to -1 then it will never execute, and is no longer in control. Perhaps that should be added to the Wiki. Perhaps you will add it for me?

dg2k:
Oh! yes of course, -1 disables it. I should have remembered this as we discussed about this a while back but in a different context.

Sure, will have a loot at the Wiki.

JerryNixon:
I have added your event, and more. 

You can see it here: https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs#L290

Issue #959, opened 2016-05-13
Adding a ButtonType=Literal button in HamburgerMenu breaks all other buttons


johanlindfors:
I'm surprised that this happens...

If I add a button into Secondary Buttons, such as the following:

```
        <Controls:HamburgerButtonInfo ButtonType="Literal">
                <Grid Background="Red" Height="96"/>
        </Controls:HamburgerButtonInfo>
```

All other buttons will cease to function, they will trigger changes on the HamburgerMenu, but no Navigation will occur...

Bug or am I missing something?

Oh, and this is in the latest preview just pulled from Nuget, and also verified in the Hamburger template project on the master branch on GitHub.

dg2k:
I've just found a bug @johanlindfors as an event is missing from Literal button DataTemplate
 See PR #962.

johanlindfors:
I hope this fix makes it into the 1.1.11

PR #962, opened 2016-05-13
Adds Loaded event to Literal Button's DataTemplate ContentPresenter.

dg2k:
In response to Issue #959 flagged by @johanlindfors.


johanlindfors:
Seems to fix the issue just fine! Great work @dg2k!

JerryNixon:
Nice. What was the root cause?

JerryNixon:
This is closed only because I added it explicitly and the code was not compatible.

Issue #966, opened 2016-05-15
Bad layout when opening menu in Hamburger project


pekspro:
If you start the Hamburger template project in a small window and then open the menu it looks like this:

![badburger](https://cloud.githubusercontent.com/assets/3163293/15272992/24c965d8-1a8b-11e6-876b-324ea3590907.png)

If you make the window larger and then small again it looks alright.

Tested with current source code.

JerryNixon:
Hey @dg2k, I think you will be able to spot the issue around this with me. Since DisplayMode, et al, are now two-way bindings, there' something happening with timing. Any bright ideas?

dg2k:
@JerryNixon - I know what it's. PR coming up (may not be bright idea but works).

dg2k:
@pekspro -- The following may help you go past the Inline DisplayMode problem. In Shell.xaml.cs ...

```
public Shell()
{
    Instance = this;
    InitializeComponent();
    ...
    Loaded += Shell_Loaded;
}

private void Shell_Loaded(object sender, RoutedEventArgs e)
{
    //  set to display mode other than Inline, e.g., CompactOverlay below
    HamburgerMenu.DisplayMode = SplitViewDisplayMode.CompactOverlay;
}
```

Just in case it helps ...

pekspro:
I just downloaded the latest source and now the menu button doesn’t open anything in the Hamburger project. I added your code also @dg2k but still the same.

Either way, personally I don’t use the hamburger yet in my projects to no issue for – yet :)

dg2k:
I can only comment after the PR is merged but before that difficult in what state you'd find my code in yours.

I noticed a similar problem yesterday (the stack-overflow problem which freezes HamburgerMenu) and PR #978 solved the problem. Let's wait and see.

PR #971, opened 2016-05-16
Adjusts PageHeader's left Margin for Inline DisplayMode

dg2k:
Re: #966 @JerryNixon 

Things go a bit haywire when DisplayMode is set to **Inline** on startup -- otherwise you may not notice this issue (but simply a problem hiding to show up later). The **Inline** DisplayMode presents a non-trivial problem -- unlike the Overlay display mode, when menu pane opens, it pushes the content (the Overlay, in contrast, simply casts the menu **over** the content), which means there is left-margin adjustment to be made for **PageHeader** when in Inline display mode. Sounds simple but there are many things at work here -- the display modes (4 of them!), the HB button container adjustment, together with PageHeader margin adjustment,  make the whole thing rather confusing.

There must be a simple XAML trick but didn't get that spark for a XAML trick so it's code-behind.  The code-behind can do the dirty job but until there's a consensus that there is no better alternative I won't be happy with this, but here we go ...

Problem: For some reason the ViewStates in the RelativePanel do not seem to trigger on startup - the reason why the startup is not as expected and extra code used to correct this (interestingly without the workaround the imperfection that shows up, such as page title positioning, etc, is self-correcting on page resize, display mode change etc, but ...).

Sufficiently commented with an indication of the VSM part that needs sorting out. It works.

**Note** It doesn't mean using Inline is imposed on the user in any way -- this is simply to accommodate starting up with Inline if the user wishes to, such a case flagged in #966.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
@@ -268,11 +268,27 @@ namespace Template10.Controls
         public SplitViewDisplayMode DisplayMode
         {
             get { return (SplitViewDisplayMode)GetValue(DisplayModeProperty); }
-            set { SetValue(DisplayModeProperty, value); }
+            set
+            {
+                // Override DisplayMode for small screen sizes.
+                // Developer should set VisualStateNormalMinWidth value (default 521d) for HamburgerMenu,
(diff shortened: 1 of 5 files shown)


Viachaslau-Zinkevich:
@dg2k could you please check if my PR fixed the issue? At least for my #972 it seems to be a clean fix. Would be nice if it fixed this one, too. At least I couldn't reproduce after that.
Oh..I'm late

dg2k:
@Viachaslau-Zinkevich  - Will have a look.

PR #974, opened 2016-05-17
Add an option "ClearCache" to HamburgerButtonInfo

schuettecarsten:
This pull request will add an option "ClearCache" to HamburgerButtonInfo. If set to true, navigation cache will be cleared when navigating to the page.


dg2k:
A pedantic point:

Would be nice to name the **ClearCache** DP as **ClearNavCache** or **ClearNavigationCache**.

JerryNixon:
@dg2k agreed. Wish we would have done that.

Issue #972, opened 2016-05-16
HamburgerMenu DisplayMode by default is Inline


Viachaslau-Zinkevich:
Hi, I fetched latest git version and got the following issue:
When app starts it's for some reason doesn't take into account any Narrow/Medium/Wide width and just uses Inline. 
When I at least once change style to any other (Narrow/Medium) -- then it starts behaving normally.

dg2k:
PR #971 will correct this, I think.

Viachaslau-Zinkevich:
@dg2k thanks, I'll check when merged. Is it also due to not triggering VisualStateGroup_CurrentStateChanged on start?

dg2k:
VisualStateGroup_CurrentStateChanged too but there is more.
In principle, the developer/user should be able to choose the appropriate display mode. There was no particular reason for setting the initial to Inline, except that it was set to be overridden by the visual state. When the visual state failed, the Inline problem cropped up. It's funny that Inline did not even feature as a candidate mode (see throughout the code and Inline doesn't show up as a display mode of interest). 

I personally like **CompactOverlay** for screen size >= **VisualStateNormalMinWidth** (same as the **News** app on **Windows 10**). T10 automatically sets to CompactOverlay for screen size >= VisualStateNormalMinWidth and < VisualStateWideMinWidth at which point it moves to the CompactInline display mode. Of course, for small screens ( < VisualStateNormalMinWidth ) **Overlay** is the one we can't do without and in my PR I even made this setting to override any other setting (and hope @JerryNixon will approve this).

For screen size >= VisualStateNormalMinWidth I think the developer (perhaps user??) should be able to choose any of the display modes. While this can be done, there is more room to make this much easier. Think of _preferred display modes_ that you can set to influence how the visual states make transition.

dg2k:
_Update:_  @Viachaslau-Zinkevich -- Set the display mode in **Shell.xaml.cs**'s _Loaded_ event handler and the problem with the Inline DisplayMode will not affect you.

```
public Shell()
{
    Instance = this;
    InitializeComponent();
    ...
    Loaded += Shell_Loaded;
}

private void Shell_Loaded(object sender, RoutedEventArgs e)
{
    //  set to display mode other than Inline, e.g., CompactOverlay below
    HamburgerMenu.DisplayMode = SplitViewDisplayMode.CompactOverlay;
}
```

Viachaslau-Zinkevich:
[...]

PR #975, opened 2016-05-17
Changed the way DisplayMode is set: in xaml now. #972

Viachaslau-Zinkevich:
I'be changed the way DisplayMode is set: added setters in XAML and removed codeBehind. This way DisplayMode will be correct from the beginning. It also seems to remove glitch from  #971.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml
@@ -268,29 +268,37 @@
 
         <VisualStateManager.VisualStateGroups>
 
-            <VisualStateGroup x:Name="VisualStateGroup" CurrentStateChanged="VisualStateGroup_CurrentStateChanged">
+            <VisualStateGroup x:Name="VisualStateGroup">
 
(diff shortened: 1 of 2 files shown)


dg2k:
@Viachaslau-Zinkevich - the comment you've asked -- can't make the comment where you started as that PR is now closed. What you suggest here will work (haven't tested) but puts us back to where we were a few days ago. @JerryNixon changed the xaml to code-behind (must be for a reason) so I would avoid going in circle. Also, you may have noticed that the latest commit by @JerryNixon has significant changes (so much so that your PR flags as having conflict) so would rather wait and closely see the latest changes in the code (which btw may solve all our problems ;-)

Viachaslau-Zinkevich:
@dg2k ye, sure, may be that was for a reason and is absolutely unnecessary in xaml. Will wait for additional comments.

Viachaslau-Zinkevich:
@JerryNixon would you please tell some details about the reason moving from XAML VisualStates to code-behind event?

PR #979, opened 2016-05-18
Avoid new navigation to the current page in some race conditions

mitkodi:
In some cases new navigation to the current page could happen.
For example if:
- On startup we navigate to Page1.
- NavigationService.CurrentPageType becomes Page1.
- NavigationService.NavigateToAsync (which updates LastNavigationType/Parameter) is called on another thread.
- Before NavigateToAsync updates LastNavigationType/Parameter, HamburgerMenu calls (after loading all nav buttons) HighlightCorrectButton, which uses CurrentPageType/Parameter to set selected button, that calls NavigationService.NavigateAsync.
- If at that time LastNavigationPage/Parameter are not updated yet, then NavigationService.NavigateAsync will call FrameFacadeInternal.Navigate to Page1.
  
  I've faced it (in more than 30%) in an app that uses protocol activation.

### DIFF
--- a/Template10 (Library)/Services/NavigationService/NavigationService.cs
+++ b/Template10 (Library)/Services/NavigationService/NavigationService.cs
@@ -222,11 +222,14 @@ namespace Template10.Services.NavigationService
             if (page == null)
                 throw new ArgumentNullException(nameof(page));
 
-            if ((page.FullName == LastNavigationType) && (parameter == LastNavigationParameter))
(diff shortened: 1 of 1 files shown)


JerryNixon:
I agree. But is the solution what you are suggesting or should we fix the timing?

mitkodi:
Since I don't see a reason to not relay on CurrentPageType/Parameter when checking whether navigation should happen, I think it's better than fixing all places where timing issues could come (now or in the future).

But of course I could be wrong :(

PR #981, opened 2016-05-18
Fix HamburgerMenu blinking when navigating

schuettecarsten:


JerryNixon:
I think you are solving a problem that does not exist, Carsten.

schuettecarsten:
Oh, it does in our solution. Never mind, I will add patch to our project fork.

PR #983, opened 2016-05-18
VisualState[Narrow|Normal|Wide]DisplayMode invoke UpdateVisualStates(…

dg2k:
~~Removes a circular call to UpdateVisualStates();~~ VisualState[Narrow|Normal|Wide]DisplayMode invoke UpdateVisualStates() on change.

Did PR again due to merge conflict being flagged. It appears that the _circular call_ problem is done manually since??

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
@@ -334,6 +334,7 @@ namespace Template10.Controls
             DependencyProperty.Register(nameof(VisualStateNarrowDisplayMode), typeof(SplitViewDisplayMode),
                 typeof(HamburgerMenu), new PropertyMetadata(SplitViewDisplayMode.Overlay, (d, e) =>
                 {
+                    (d as HamburgerMenu).UpdateVisualStates();
                     WriteDebug(nameof(VisualStateNarrowDisplayMode), e);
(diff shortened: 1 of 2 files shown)


dg2k:
@JerryNixon -- any luck with this PR?
If setting VisualState[Narrow|Normal|Wide]DisplayMode doesn't take immediate effect (by calling UpdateVisualStates()) it's not working as intended. It would be nice if this is part of the new release. I also see no reason for an event handler associated with this (other than for the sake of symmetry i.e., making them similar to other dependency properties such as DisplayMode, et cetera).

Template10 (Library)/Controls/HamburgerMenu.PublicProperties.xaml.cs
```diff
@@ -334,6 +334,7 @@ public SplitViewDisplayMode VisualStateNarrowDisplayMode
             DependencyProperty.Register(nameof(VisualStateNarrowDisplayMode), typeof(SplitViewDisplayMode),
                 typeof(HamburgerMenu), new PropertyMetadata(SplitViewDisplayMode.Overlay, (d, e) =>
                 {
+                    (d as HamburgerMenu).UpdateVisualStates();
```
JerryNixon, line 337: Let's change this so there is no logic in this file. Use the InternalChanged methods to add functionality in the Ham.xaml.cs file. Okay?

dg2k:
Sure. Will amend that.

PR #978, opened 2016-05-18
Removes a circular call to UpdateVisualStates(); VisualState[Narrow|N…

dg2k:
StackOverflow!
I don't mean _The Website_ but there's a circular call to `UpdateVisualStates()` causing stackoverflow.

UpdateVisualStates() -> HamburgerMenu_DisplayModeChanged() -> UpdateControl() -> ~~if (!IsFullScreen) UpdateVisualStates()~~

Also, **VisualState[Narrow|Normal|Wide]DisplayMode** now invoke UpdateVisualStates() upon change as they should.

PR #985, opened 2016-05-18
Adds FirstDescendant extension to complement similar utilities. To be…

dg2k:
Adds **FirstDescendant** extension to complement similar extensions in the **XamlUtils.cs**. To be used for a sample project.

### DIFF
--- a/Template10 (Library)/Utils/XamlUtils.cs
+++ b/Template10 (Library)/Utils/XamlUtils.cs
@@ -109,6 +109,37 @@ namespace Template10.Utils
             }
         }
 
+        public static T FirstDescendant<T>(this DependencyObject control) where T : DependencyObject
+        {
+            return control.Descendants().OfType<T>().FirstOrDefault();
+        }
+
+        public static IEnumerable<DependencyObject> Descendants(this DependencyObject control)
+        {
+            var queue = new Queue<DependencyObject>();
+            var count = VisualTreeHelper.GetChildrenCount(control);
+
+            for (int i = 0; i < count; i++)
+            {
+                var child = VisualTreeHelper.GetChild(control, i);
+                yield return child;
+                queue.Enqueue(child);
+            }
+
+            while (queue.Count > 0)
+            {
+                var parent = queue.Dequeue();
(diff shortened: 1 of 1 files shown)


JerryNixon:
There's already a `FirstChild<T>()`, why are you adding this?

dg2k:
Wasn't there when I last checked; must be a very recent addition.

Saying that, FirstChild and FirstDescendant may be different (depends on the logic of the definition). You may not have a control type as a child, but you can have it as a grand-child or even further down. FirstDescendant is just the first that is found regardless of where. I have to look into how FirstChild is defined; if you still call it a child on step below the immediate hierarchy on the visual tree, perhaps it needs to be renamed FirstDescendant. The same logic goes for parent and ancestor.

JerryNixon:
Give me an example Visual Tree that would do what you are saying. Control > Not Control > Control.

dg2k:
Start from Grid and drill down until you find a FontIcon or something. I am not interested in the intervening controls (which may be a stackpanel, etc).

JerryNixon:
FindChild can't do that?

dg2k:
Two things:
A quick inspection of FindChild shows that it is not an extension - I have to reference it as `XamlUtils.FirstChild <T>(control)` rather than `control.FirstChild <T>()`.

Secondly, I have to test if FirstChild gives me the same result as ~~FirstAncestor~~ FirstDescendant. If it does, perhaps it needs to be renamed. Just my opinion based on how I have been used to the terminology.

JerryNixon:
I get it. Let's just fix FirstChild(). Make sense?

dg2k:
Sure, no problem.

JerryNixon:
All of these could be duplicated as extension methods, for sure.

dg2k:
Yes, `this` does all the magic of course.
Interestingly, I posted PR for FirstDescendant over [a day ago](https://github.com/Windows-XAML/Template10/pull/978/commits/e1d070683623400a6e6c2357fcf181f859f543f5). FirstChild is in existence only for 20 hours according to the merge log. Just to confirm my comment above.

JerryNixon:
Yep. You are an inspiration.

PR #994, opened 2016-05-20
Make sure OnNavigatingFrom and OnNavigatedFrom use same page/viewmodel

schuettecarsten:
Make sure that OnNavigatingFrom and OnNavigatedFrom are called on the same Page/ViewModel.

It might happen (always happens here) that FrameFacadeInternal.Content changes between call of NavigatingFromAsync and NavigateFromAsync. In old implementation, this causes that NavigateFromAsync is called on the page where user just navigated to. The PR fixes this by using the same page/viewmodel instance for NavigatingFromAsync and NavigateFromAsync.

Unfortunately, the changes are quick and dirty and maybe they can be improved.

### DIFF
--- a/Template10 (Library)/Services/NavigationService/NavigationService.cs
+++ b/Template10 (Library)/Services/NavigationService/NavigationService.cs
@@ -56,11 +56,23 @@ namespace Template10.Services.NavigationService
                 if (e.Suspending)
                     return;
 
-                // allow the viewmodel to cancel navigation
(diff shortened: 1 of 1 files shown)


JerryNixon:
No, Carsten, we can't change the signature of INavigable. I agree with the goal.
But, you need to dream up a different way.

schuettecarsten:
I did not change the signature for INavigable. The changes are only for the private methods in NavigationService. Please take a look into the changes, only one file changed, no public interface changes.

PR #996, opened 2016-05-20
Changes FirstChild method to an Extension.

dg2k:
No biggy PR but has to be done.


JerryNixon:
I will merge this but I will also update it slightly.

PR #998, opened 2016-05-20
Sample project showing runtime setting of themes defined in Styles/Custom.xaml.

dg2k:
Sample project showing run-time setting of themes defined in _Styles/Custom.xaml_.

It's easy to do it on startup (the style def does it for you), but not so easy if you want to change the themes during run-time (say, if you'd like to provide user option to change themes and colors of your app).

In short, theme setting during run-time has some interesting challenges and this sample project shows how to do it.


JerryNixon:
This is dramatically better than the theme setting in the Templates?

dg2k:
@JerryNixon -- the background to this has been discussed in some detail. It is not suggested as "better" at all; this is addressing a different problem -- run-time setting of themes.

Please refer to a detailed commentary supported by gif videos in PR #897; see also related comments in #887.
- PR #897 is responsible for the[ `public void RefreshVisualState()`](https://github.com/Windows-XAML/Template10/blob/master/Template10%20%28Library%29/Controls/HamburgerMenu.xaml.cs#L699) method to be implemented in HamburgerMenu.xaml.cs.
- Without the Visual State correction and refresh (discussed in #897) and later implemented with **RefreshVisualState()** in code-behind, the xaml style alone will not do the rendering of Hamburgermenu as desired during runtime. The problem boils down to the Visual State not refreshing the HamburgerMenu style components (such as _NavButtonCheckedBackground_, _NavButtonCheckedForeground_, etc).
- Hence, why this Sample Project is relevant to demonstrate how to set themes at run-time by pulling together the different strands of the solution to tackle the Visual State problem.
- Hopefully this problem may not be there with a forthcoming SDK update (assuming the visual state anomaly referred to here is taken care of)  but from experience immediate resolution of such problem with a new SDK is not always the case.

PR #1011, opened 2016-05-25
Ensures that change in VisualState[Narrow|Normal|Wide]DisplayMode wil…

dg2k:
- Ensures that changes in **VisualState[Narrow|Normal|Wide]DisplayMode** properties will have immediate effect.
- This is a re-submission of #983 with an amendment and with changes now confined to a single file.


JerryNixon:
Nice and clean

PR #1018, opened 2016-05-28
Makes app resume after task switch.

pekspro:
Currently Template10 only resumes app if it was activated when it was activated from primary tile. However, it should also resume if it gets activated via task switch. This is detected by checking if TileId is an empty string. Reference:

https://msdn.microsoft.com/en-us/library/windows/apps/hh700588(v=win.10).aspx?appid=dev14idef1&l=en-us&k=k(windows.applicationmodel.activation.ilaunchactivatedeventargs.tileid)%3bk(solutionitemsproject)%3bk(targetframeworkmoniker-.netcore,version%3dv5.0)%3bk(devlang-csharp)&rd=true

I detected this when running my app on Windows Phone, and these changes seems to solve the issue. I haven’t been able to replicated this on a computer. And I have no idea how to test this in Visual Studio :-/.

### DIFF
--- a/Template10 (Library)/Common/BootStrapper.cs
+++ b/Template10 (Library)/Common/BootStrapper.cs
@@ -213,10 +213,11 @@ namespace Template10.Common
                     {
                         OnResuming(this, null, AppExecutionState.Terminated);
 
-                        /*
+						/*
                             Restore state if you need to/can do.
-                            Remember that only the primary tile should restore.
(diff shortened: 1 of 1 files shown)


schuettecarsten:
Please check spaces/tabs. Template10 uses spaces.

pekspro:
OK, that's done.

Issue #1014, opened 2016-05-26
Application theme is always light when application starts.


aannenko:
Application theme is always light when application starts. This is true even when you switch to dark theme in the app's settings and restart the app.
To fix this, simply remove the `RequestedTheme="Light"` property setter from the App.xaml file.

calloncampbell:
I believe the reason it's defaulted to white is because that is the new standard for Microsoft's UWP apps.

dg2k:
Don't think white is the preferred theme although apps tend to choose more white than dark.
The three themes, namely Default (aka Dark) Theme, Light Theme and HighContrast Theme are the available standards. There is no suggestion of a preferred standard. It's down to the developer and the user.

The developer may provide options for the user to choose a preferred theme through settings; in some cases the developer may for whatever reason restrict the options. If on the other hand the app is aimed at where user accessibility is paramount (say, for visually impaired) then the implementation of HighContrast Theme may be a requirement.

To come back to the `RequestedTheme="Light"` in the App.xaml file -- that is not a suggestion for favoring the Light Theme. You just got to pick one to start off, it could have been equally `RequestedTheme="Default"` and nothing is implied by that.

In the final implementation,
- the theme is set in the App ctor(). The theme value normally comes from settings that the user persisted through options setting.
- the developer conveniently defines theme options via a style file (Custom.xaml used for sample projects). The style definitions can be applied to various controls as Static Resources and this can be as comprehensive as one wishes - not only for HamburgerMenu and PageHeader, but for theme styling specific to the app.
- you can even change the themes and colors at runtime ... where do you want to stop ? It all depends ...

aannenko:
Cool! But this time it is just a small defect. The theme is initialized twice: once in the App.xaml.cs ctor and the second time - in the Custom.xaml
If the Custom.xaml does not contain a theme setter and settings storage doesn't have a key for a current theme, the SettingsService instance's AppTheme property returns light theme which is just enough for not having a theme setter in the Custom.xaml

JerryNixon:
Worth pointing out that Dark is the internally preferred theme only on phone, Light is the preferred theme on desktop applications. You can confirm this with the first-party apps shipping with Windows 10. Or, you can just trust me. This, by the way, was a decision that was made recently. In the recent past, Dark was the default theme for all apps. I've noticed that many of the bigger first party apps, like News, still seem to prefer Light even on mobile.

jp-weber:
[...]

PR #1035, opened 2016-06-05
Better theme control (specially for the redstone release)

jp-weber:
I have implemented a checkbox and two other variable in hamburger sample, in order to make the control of the theme better and ensures that the correct system theme is taken at start. 

Especially for Windows RS1 it is this change logical then.
#1014


JerryNixon:
I know this is just sitting here, but we'll do something like it in Redstone if not this code exactly.

PR #1039, opened 2016-06-08
Corrects a method name for KeyBehavior CallMethodAction in Search Sample Project.

dg2k:
You'll notice this little bug only if you press the **Enter** key,  but overlooked for so long because we're more inclined to tap on the search **Button** for same action. The **CallMethodAction** was not provided with the correct method name hence an Exception [here](https://github.com/Windows-XAML/Template10/blob/master/Template10%20%28Library%29/Behaviors/TextBoxEnterKeyBehavior.cs#L34).

This PR reminded me that **TextBoxEnterKeyBehavior** needs to be updated with **KeyBehavior** so more PR's coming up.


dg2k:
Replaced with PR #1046 due to Branch Conflict.

PR #1045, opened 2016-06-10
Fixes "The name 'PropertyChanged_ViewModel' does not exist in the current context"

shahthepro:
A fix for the error "The name 'PropertyChanged_ViewModel' does not exist in the current context" that appears while compiling a newly created UWP project with Minimal or Hamburger templates. 

Going by the [docs here](https://msdn.microsoft.com/en-us/library/system.windows.data.binding.mode%28v=vs.110%29.aspx), The BindingMode of TextBlock element defaults to OneWay and so there is no need to specify it explicitly.


JerryNixon:
Leave a comment

PR #1056, opened 2016-06-13
Rate and Review Service

AparnaChinya:
Added a service which pops up the dialog to Rate and Review the app which is published on store, after the user uses the app certain number of times.


JerryNixon:
I am going to get #1070 merged in, and then I want this to be part of that solution.
Make sense? Give him a few days to update and merge.

PR #1065, opened 2016-06-18
Synchronous Navigate methods should wait until navigation finishes

schuettecarsten:
Synchronous Navigate methods should wait until navigation finishes. So with PR removes async keyword from syncronous Navigate methods and lets them wait using GetAwaiter().GetResult() call. The ConfigureAwait(false) makes sure that this does not block the UI thread.


JerryNixon:
Why does this matter?

JerryNixon:
Can you update this?

schuettecarsten:
Will update on Tuesday, I'm on holidays :)

schuettecarsten:
Updated. Happy merging :)

Issue #1069, opened 2016-06-19
Add service to ask user to review the app after a certain amount of usage


dkackman:
Many apps will ask the user to leave a review in the app store after a certain period. Add a component that will allow an app to register to receive a "Please review this app" nag dialog after a certain number of launches and/or period of time since the first launch.

PR #1070, opened 2016-06-19
AppNagService

dkackman:
#1069

Service that allows an app to register to nag the user to review the app after a certain period of time or launches.

A couple notes on the code:
- I put it in the `Template10.Services.MarketplaceService`namespace as that seemed appropriate even though most other services seem to have a 1:1 between namespace and service
- `AppNagService`took dependencies on `IDialogService`, `IFileService`and `IMarketplaceService`. I don't see other examples of inter-service dependencies so not sure what patterns you prefer there. Those dependencies can be passed to the service constructor as interfaces so it should support IoC.

I did create a sample app, not included in this pull request, that I can create a PR for if this gets merged.


JerryNixon:
Cool. Let's remove "Nag" from it, and the files. And let's make the service Sealed. 
Good work.

JerryNixon:
Looking closer. Let's make IMarketplaceService and all the other files nested INSIDE the rot MarketplaceService namespace.

dkackman:
> Cool. Let's remove "Nag" from it, and the files. 

How does `AppReviewService`sound? 

> Looking closer. Let's make IMarketplaceService and all the other files nested INSIDE the rot MarketplaceService namespace. 

Can you clarify that? Not sure what you are requesting.

JerryNixon:
Let's do an about-face here. Let's call it NagService. Let make it reusable, though. Let's make it so, if your app has an updagrade and you want the user to upgrade, use the NagService. If your app has new data to download, use the NagService. And, if your app needs reviw, use the NagService. Just like you have it, but the key (for one) will need to change and be customizable (somehow). This make sense? A totally reusable NagService for any use case.

Template10.Services.NagService.INagService;
[sealed] Template10.Services.NagService.NagService : INagService;

Then we also need a MarketplaceService, almost 100% of what you already have. This will use the NagService. Which, I realize s almost where you were originally heading. 

Template10.Services.MarketplaceService.IMarketplaceService;
[sealed] Template10.Services.MarketplaceService.MarketplaceService : IMarketplaceService;

Here's a note: a helper, fundamentally, can be used in any project. A service is specific, and might need to be re-implemented. A service uses the helper. The helper is just a simplification of another API or set of API. WHy do we have a helper? To simplify a complex implementation - but it is NEVER specific to a project. A service, on the otherhand, is always specific. If it is generic and reusable, that's jsut a coincidence and a bonus. It's also why it is sealed!

dkackman:
Was thinking along these lines originally so have some ideas on how to make it work.

JerryNixon:
Maybe a napkin drawing before you code it all up?

dkackman:
Here's the basic sketch in some roughed out stubs:

This class describes the nag and is how user code sets up a specific nag reminder:

```
public class Nag
{
    public string Id { get; set; }

    public bool AllowDefer { get; set; }

    public string MessageTemplate { get; set; }

    public string NoButtonText { get; set; } = "No, thanks";

    public string TitleTemplate { get; set; }

    public string YesButtonText { get; set; } = "Yes";

    public string DeferButtonText { get; set; } = "Not now";

    public Action NagAction { get; set; }
}
```

`Id`is how nag state would be indexed on the filesystem in order to manage the user's response, or lack thereof, to a nag e.g.  `"Template10.Services.NagService.{Id}.json"`

Two thoughts on how to specify Id: 1) Let the user specify the id and make sure their app's nag ids are unique. 2) SHA1 hash of the `MessageTemplate`. I'm leaning toward the hash as it is less for user code to manage, and the message template really is the unique element of a particular nag within the scope of an app. 

There might be the need to allow user code to persist state with the nag to handle cases like: you've nagged the user to upgrade before, and they responded, but now there is a new upgrade. How does the app know to ignore the previous response and nag again? Squirreling away the last version nagged about with the response state is one way an app could manage that sort of scenario. I haven't thought that all the way through yet though.

`INagService`looks something like this, probably with some additions to allow nag responses to be inspected, reset or deleted:

```
public interface INagService
{
    Task Register(Nag nag, int launches, TimeSpan duration);

    Task Register(Nag nag, int launches);

    Task Register(Nag nag, TimeSpan duration);
}
```

The implementation of `INagService`is essentially the current `AppReviewNagService`code extended to support 1-n nags and associated housekeeping.

In regards to your point about Service vs Helper: though I am picking up what you're laying down, I'm not sure what changes in current PR code structure would better fit that approach. Is it that the `AppReviewNagService`has those 3 service dependencies and should avoid those in favor of instantiating the helper classes for file system/dialogs/marketplace internal to the service implementation? This would make the service project specific and it wouldn't need to worry about dependency management, but I'm not positive that is the direction you were heading. If there is a specific example that would help this code better fit the design philosophy of the project just let me know.

JerryNixon:
Looks great. Prob need to make Id a `private set` and pass it in the constructor to force the developer to use it properly. Otherwise, I think I would write it like this. Thanks for the up front effort here. ANd, yeah, I think Nag is an okay term here.

dkackman:
Cool. Should be able to get to it after the 4th. I'm thinking we should close this PR and I'll open a new one with the refactored implementation.

Issue #1071, opened 2016-06-19
FileService namespace exists in both Library and Services assemblies


dkackman:
`FileHelper`and `StorageStrategies`types exist in the same namespace in both Template10 (Services) and Template10 (Library).

Looks like it could either be moved entirely into the Library assembly or the Services assembly, depending on what the ultimate design intent is. Nothing in Library seems to have a dependency on either type.

I can make the change and submit a pull request but would need some guidance as to which assembly it should be in.

PR #1073, opened 2016-06-20
https://github.com/Windows-XAML/Template10/issues/1071

dkackman:
#1071 Remove namesspace and types for FileSystemSErvice which also exist in the Services assembly


JerryNixon:
This is backwards. We're in the slow process of moving services into the library. This belongs there and does not belong in the Service assembly anymore. The ultimate goal is that the core services will be in the library. It's worth pointing out that in our next version, they will be separate assemblies to allow develoers to keep their apps as min as poss. In the meantime, for v12 we're putting it in. Thanks for noticing that conflict with Services.

dkackman:
Makes sense. Is services going to eventually take a dependency on library? The reason I ask is that the AppReview/Nag service has a dependency on the IFileSystemService stuff in services. So once file system moves, the app review service will need to move as well or services will need a reference to library, which it currently doesn't have.

I can easily do a PR with file system stuff removed from services if you're cool with it having a reference to library. If not I may need to rethink how the app review service persists state for fear of horking up your longer term plan.

PR #1086, opened 2016-06-28
NetworkAvailableService

aline-almeida:
- add NetworkAvailableService to library
- make AvailabilityChanged action available in service

### DIFF
--- a/Template10 (Library)/Services/NetworkAvailableService/ConnectionTypes.cs
+++ b/Template10 (Library)/Services/NetworkAvailableService/ConnectionTypes.cs
@@ -1,0 +1,9 @@
+﻿namespace Template10.Services.NetworkAvailableService
+{
+    public enum ConnectionTypes
+    {
+        None,
+        LocalNetwork,
+        Internet
+    }
+}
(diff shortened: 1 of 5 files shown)


JerryNixon:
Nice work @aline-almeida. This is a beautiful job of coding. Like it came out of my own fingers. Thank you for taking the time to not only do good work but also match the structure of Template 10 internally. 

The only issue here is that there are two things in one and we need to separate them. There is a difference between IsNetworkAvailable and IsInternetAvailable. Connected to an available internet is a valuable thing to know. Having internet access is also valuable. We need to know both.  

Does this make sense?

``` csharp
enum Types { None, Network, Internet }
Action<Types> AvailabilityChanged { get; set; } 
```

Also

``` csharp
Task<bool> IsInternetAvailable();
Task<bool> IsNetworkAvailable();
```

That should do it, though it might take some tweaking.

aline-almeida:
Hi Jerry!
In fact, the code is really yours, I've just included it in the right place and exposed the action :)
Your suggestion makes sense. Done! Hope it meets your requirements.

PR #1085, opened 2016-06-28
Added NavigationTransitionInfo override to NavigationService.GoBack()

ghost:
this is in response to Issue #1074 
please check if I have done anything wrong.

PR #1087, opened 2016-06-29
Added optional NavigationTransitionInfo parameter to NavigationService.GoBack()

ghost:
#1085


JerryNixon:
And should we do Forward() before we wrap this up?

ghost:
the built in `Frame.GoForward()`does not take `NavigationTransitionInfo` as an argument though

PR #1088, opened 2016-06-29
NavToPageAction doesn't work.

GFlisch:
Hi.

Navigate expects a page type to navigate on a page. The Page dependency property which is an object must be bound in xaml to a page type (System.Type).

The pagetype variable (returned from metadataProvider.GetXamlType(Page as Type); ) when cast to  Type is null...

Just giving the Page to the navigation is enough.

### DIFF
--- a/Template10 (Library)/Behaviors/NavToPageAction.cs
+++ b/Template10 (Library)/Behaviors/NavToPageAction.cs
@@ -26,15 +26,7 @@ namespace Template10.Behaviors
             if (nav == null)
                 throw new NullReferenceException($"Cannot find NavigationService for {Frame.ToString()}.");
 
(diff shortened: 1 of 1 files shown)


JerryNixon:
Does this really work? (Would test, but I am traveling)

JerryNixon:
Is this related? https://github.com/Windows-XAML/Template10/issues/1093

GFlisch:
Hi Jerry,
I have a sample project (Minimal Template).
In the MainPage, the Key behaviour="Enter" will now trigger a TestAction wich is a copy of the proposal I did. So when you enter in the input text you navigate to the DetailPage.
https://1drv.ms/u/s!AlpqJ8sjLiUwr-wHMYyat93hk37dyg
Gilles

Issue #1082, opened 2016-06-27
Massive performance issues because of LayoutUpdated event handlers


Opiumtm:
There are some handlers to [LayoutUpdated](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.frameworkelement.layoutupdated) event in Template10 code.

Quote from MSDN:

> When you handle LayoutUpdated, do not rely on the sender value. For LayoutUpdated, sender is always null, regardless of where the handler is attached. This is to prevent handlers from assigning any meaning to sender, such as implying that it was that specific element that fired the event out of the visual tree. LayoutUpdated implies that something in the overall visual tree has changed, and each specific object anywhere in the tree has the option of handling this occurrence. If you're familiar with lower-level render API design, you can equate LayoutUpdated being fired as similar to a "redraw needed" flag being set as part of an object-driven, retained-mode rendering logic.
> 
> Because LayoutUpdated fires in many circumstances and isn't always specific to an object that actually changes, consider whether handling the SizeChanged event instead is more appropriate for your scenario.

I already pointed that hook on [LayoutUpdated](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.frameworkelement.layoutupdated) is a very dangerous thing from performance perspective. Microsoft itself discourage to use this event in many scenarios, it can be used only as a last resort and with extreme care as it will lead to performance degradation.

I already have asked why some parts of code (global-scale behaviors, first to mention) handle this dangerous event. No answer was provided.

As I already pointed, these event handlers should be removed, but we must understand why these handlers to [LayoutUpdated](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.frameworkelement.layoutupdated) event was attached at first place, what's the reason?

Opiumtm:
Some memory refreshment: #902

I provided quick-fix with #906 pull request (was accepted).
But that was only a partial solution which removed most blatant performance degradation but doesn't fix core issue which is a handling of `LayoutUpdated` event. I mentioned that core issue is with this event and doesn't removed these handlers in pull request because I don't know why these handlers are created at first place. No information available about reasons of this questionable and problematic technical decision.

JerryNixon:
I understand, but I thought this was already resolved. Did it get un-resolved?

Opiumtm:
[...]

PR #1092, opened 2016-06-30
Fix for  #1082 issue (removed LayoutUpdated handlers)

Opiumtm:
Proposed fix for #1082 issue

@JerryNixon Please, review these changes carefully

I have made changes as I understand its logic, because no one helped me with information. If I'm wrong, please, point me on it.
### EllipsisBehavior

Simply removed `LayoutUpdated` event handler  because `Loaded` event handling is enough. `Loaded` event on WinRT/UWP is guaranteed to be fired after final visual tree construction. There was event race condition in Silverlight implementation of Xaml UI, but no such issues on WinRT/UWP (according to MSDN)
### HamburgerMenu

Moved these calls from `LayoutUpdated` event handler to `Loaded` event handler because it's anyway one-shot event and is needed to be called after visual tree is finally constructed. `Loaded` event handler is a safe place to do it.

``` C#
UpdateVisualStates();
UpdateControl();
```
### NavButtonBehavior

Most heavily changed class.

Fixed some bugs and potentially risky places. For example, `BootStrapper.Current.ShellBackButtonUpdated` event handler was attached in both `Attach`
 and `Detach` methods. Copy/paste without change of `+=` to `-=` in `Detach` method probably a cause 😄 
Fixed potentially incorrect behavior if `Frame` property is changed at run-time.

``` C#
        private static void FrameChanged(DependencyObject d, DependencyPropertyChangedEventArgs args)
        {
            var behavior = (NavButtonBehavior)d;
            DetachFrameEvents(behavior, args.OldValue as Frame);
            AttachFrameEvents(behavior, args.NewValue as Frame);
            behavior.CalculateThrottled();
        }
```

event handlers from old Frame should be detached or there will be memory leak 

Moved dependency property changed event registration handles into private class

``` C#
        private class FrameEventRegistration
        {
            public long GoBackReg;
            public long GoForwardReg;
        }
```

and used `ConditionalWeakTable` to bind this frame registration to `Frame` instance itself - just to be sure if we removing handlers from a correct `Frame` instance

Attached handler to `Frame.Navigated` event to update navigation buttons visibility after obvious event of frame navigation.

Attached handler to `DeviceUtils.Changed` event  to update navigation buttons visibility after change of device disposition. For example, after going to full screen window mode or to continuum mode on mobile device.

Attached handler to `Frame.Loaded` event to update navigation buttons visibility after frame visual tree initialization is complete.

As I think, handling of `Frame.Navigated`, `DeviceUtils.Changed`, `Frame.Loaded` is enough to cover all possible triggers which can require navigation buttons visibility refresh.

As a minor code refactoring moved button visibility logic into static class `NavButtonsHelper` because public static methods shouldn't be declared inside non-static class if they don't return nor accept as a parameter such non-static class instance references.

### DIFF
--- a/Template10 (Library)/Behaviors/DeviceDispositionBehavior.cs
+++ b/Template10 (Library)/Behaviors/DeviceDispositionBehavior.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Markup;
+using Template10.Utils;
 
 namespace Template10.Behaviors
 {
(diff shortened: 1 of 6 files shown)


Opiumtm:
@JerryNixon Don't merge this PR now
Found issue with `DeviceUtils`

schuettecarsten:
@Opiumtm : You're fixing more things than the LayoutUpdated handlers, don't you?

Opiumtm:
@JerryNixon @schuettecarsten 
Found no direct problems with this pull request. Mistaken **new** dire performance issues.

It's hard to measure performance improvements when codebase independently introduced new performance problems.

Investigating what commit and from who exactly introduced **new** extreme performance degradation.

Opiumtm:
@JerryNixon @schuettecarsten 
#1112 here investigation results for another introduced performance issue 😢

PR #1099, opened 2016-07-06
Update dependency to Newtonsoft.Json from 8.0.3 to 9.0.1 (latest)

schuettecarsten:
Works fine with our project (using Services/Library), compile-tested for Template10 samples only.


JerryNixon:
I have reverted this

PR #1120, opened 2016-07-15
Adds GroupName dependency property to facilitate submenu implementation

dg2k:
### Summary
- Updated version of PR #1114.
- Adds a **GroupName** dependency property to facilitate submenu implementation;
- Adds an extension to get a list of submenu NavButtons.
### In XAML

**parent NavButton  (the first in the Group and must be Command button):**

```
<Controls:HamburgerButtonInfo GroupName="groupName" ButtonType="Command" Tapped="MenuTapped" ... >
```

**submenu NavButtons (can be any button type):**

```
<Controls:HamburgerButtonInfo GroupName="groupName" ... >
```
### In Codebehind

```
private void MenuTapped(object sender, RoutedEventArgs e)
{
     var hbi = (sender as HamburgerButtonInfo);
     var subMenuButtons = hbi.ItemsInGroup();

     ... // do visibility toggling on submenus (see previous example) or other operations ...
}
```
- `hbi.ItemsInGroup();` extension returns a list of submenu items (in PrimaryButtons) as `List<HamburgerButtonInfo>`.
- `hbi.ItemsInGroup(IncludeSecondaryButtons:true);` provides an option to include SecondaryButtons.
- No need to pass on the GroupName string; the extension does that job.
- No need to exclude the parent NavButton (of type Command) from the returned List; again, the extension does that job.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerButtonInfo.cs
+++ b/Template10 (Library)/Controls/HamburgerButtonInfo.cs
@@ -69,6 +69,22 @@ namespace Template10.Controls
           typeof(HamburgerButtonInfo), new PropertyMetadata(null));
 
         /// <summary>
+        /// Sets and gets the GroupName property.
+        /// In simplest form, a NavButton acting as a parent menu handles its Tapped event with the remaining group 
+        /// members as submenu children. This parent can then act upon its children, such as toggling their visibility.
+        /// You can, no doubt, find other more advanced uses for this though haven't figured out one yet ...
+        /// 
+        /// </summary>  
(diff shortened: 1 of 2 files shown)


JerryNixon:
Great work. Good thinking to make it an extension method, too. But since there is a Template10Utils class, and since this is specific to T10 and not XAML, let's put it in the T10 Utils and not the Xaml Utils. Do you agree?

dg2k:
Agree. Will update that.
@JerryNixon -- updated as #1121.

PR #1132, opened 2016-07-18
Adds VisbilityChangeBehavior

pekspro:
I made this behavior a while ago and I found it to be very useful. It’s triggers when a component becomes visible (including all parents). I use it for instance to set focus on button or lists that becomes visible when data has been downloaded, which happens a lot in my applications :) It can also be used instead of EventTriggerBehavior to set focus when a page has been downloaded.

Not sure if this should be a part of Template10, but I make a pull request and let you decide :)

### DIFF
--- a/Template10 (Library)/Behaviors/VisbilityChangeBehavior.cs
+++ b/Template10 (Library)/Behaviors/VisbilityChangeBehavior.cs
@@ -1,0 +1,94 @@
+﻿using Microsoft.Xaml.Interactivity;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Markup;
+
+namespace Template10.Behaviors
+{
+    [ContentProperty(Name = nameof(Actions))]
+    [TypeConstraint(typeof(FrameworkElement))]
+    public class VisbilityChangeBehavior : DependencyObject, IBehavior
+    {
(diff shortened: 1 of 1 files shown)


JerryNixon:
This is very much an edge case, also it implies that when it is fired the associated object is visible, but you are not actually checking that - you are only checking visibility is set. But with render transforms, opacity, clipping boundaries, scroll viewers, overdraw, and more - this isn't actually reporting visibility. As a matter of discussion, the advertising team had the same trouble when they were creating their ad control. They needed to determine if the developer was gaming the system by calling the control but not actually showing it. I can tell you this, the logic necessary to determine if a control is actually visible is painfully complex and can dramatically impact performance. That being said, this behavior is something specfic-enough I think the developer should be responsible for it, not the framework. Okay? Last thing, I see you are attaching to LayoutChanged, but you should search that approach - generally, it's SizedChanged you want to attach to for the sake of performance, esp. memory consumption. Thanks for your PR. There's nothing wrong with your code, and I appreciate your coding style - it's similar to mine. I just think this behavior does not fit in T10. I hope you understand.

pekspro:
That’s absolutely fine :). Thanks for the extended information. Btw I tried with SizedChanged first but that didn’t work (don’t remember why) so I changed to LayoutChanged and it seems to be fine.

PR #1148, opened 2016-07-28
Add forward page type to NavigatingEventArgs

ghost:
ForwardPageType lets know the page type NavigationService is navigating
to.


JerryNixon:
Let's call it TargetPageType and let's also include TargetPageParameter, right?

ghost:
@JerryNixon Sounds good!

PR #1185, opened 2016-08-20
Fixing up the validation

ArtjomP:
ValidatableViewModelBase is Json Serializable now. 
Nuget Panteam.Template10.Validation 1.0.24 includes this update


Samples/Validation/Models/User.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Template10.Samples.ValidationSample.Models
 {
-    public class User : ModelBase
+    public class User : ValidatableViewModelBase
```
JerryNixon, line 6: Would you be willing to call this ValidatableModelBase instead?

Template10 (Validation)/Validation/ValidatableViewModelBase.cs
```diff
@@ -9,16 +10,17 @@
 
 namespace Template10.Validation
 {
-    public abstract class ModelBase : BindableBase, IModel
+    public abstract class ValidatableViewModelBase : ViewModelBase, IModel
     {
-        public ModelBase()
+        public ValidatableViewModelBase()
```
JerryNixon, line 15: Could we change ViewModelBase to ModelBase?

Samples/Validation/Models/User.cs
ArtjomP: Sure

Template10 (Validation)/Validation/ValidatableViewModelBase.cs
ArtjomP: I'm sorry I don't quite understand what do you mean. Your base class name in T10 is ViewModelBase, it's the actual information: https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Mvvm/ViewModelBase.cs . Do you plan to change the name of the base class ViewModelBase to ModelBase in the core Template10 Library?

tibitoth: I think he meant to rename the`ValidatableViewModelBase` to `ValidatableModelBase`

ArtjomP: He said this in the previous comment on the file User.cs. If it's that I don't care.

JerryNixon: This is to be the base class of a Model, right? Not a View-model. It seems to me that we would call this `ValidatableModelBase` since it is the base for a model. If it were the base for a view-model we would call it `ValidatableViewModelBase`. But, it isn't. This is for models. Right?

Issue #1178, opened 2016-08-14
Mobile Emulator 10.0.14393.0 WVGA 4 inch 1GB (Windows AU SDK)


dg2k:
I notice something strange with a 4 inch Mobile Emulator (that comes with Windows AU SDK)  as shown in the image below (PageHeader control moves down). I  do not think the problem has to do with T10, rather with the Emulator. I coudn't confirm this for sure as I do not have a 4 inch device. All the other 10.0.14393.0 emulators display correctly.

Can someone with 4 inch device (e.g., **Lumia 430**, **Lumia 435**) please confirm if **Hamburger Project** displays okay when running under VS 2015 Update 3, with project **Target version** as **Build 14393**?

![t10](https://cloud.githubusercontent.com/assets/14026786/17648816/6acf85b2-621b-11e6-8f23-559aff3387e3.PNG)

RagBuster:
I have a project which isn't using Template 10 but I am experiencing visual problems like you describe at the top and also at the bottom, looks to only effect the 4" emulator. Looks like the page content has been shifted down by 1cm where content goes below the bottom appbar.

dg2k:
Thanks for the feedback. This is in line with my suspicion that it's the 4" emulator that may be buggy. Yes, I also see the visual problem at the bottom.

dubstylee:
I noticed a similar thing with the emulators, but the issue is not there on a device. That white gap at the top is simply where the carrier name/wifi signal/etc. would be normally.

FahadMustafa:
I am seeing this issue on actual devices (Lumia 950 and Lumia 550) running T10 v1.1.11. Both devices are running Windows AU.

dg2k:
I tested on Lumia 950 XL and all is fine (**later update:** BUT the VisualState problem is there!). Not sure if resolution difference between 950 and 950 XL makes any difference.

Based on my initial test with Build 14393, I was of the opinion that it was only the 4 inch emulator that was buggy but I could be wrong.

**UPDATE**
Further to what I reported in the above, I note further issues with 14393 for Mobile. The Visual States that work with 10586 do not work with 14393 as they should. Fortunately, the test scenarios are as straightforward as switching between 10586 and 14393 emulators for the same code, and differences can be observed clearly with 14393 not behaving as expected. This is strange.

JerryNixon:
14393 support is forthcoming.

PR #1199, opened 2016-08-31
Solves PageHeader's incorrect vertical alignment on Build 14393 Platform

dg2k:
If you run your app built around T10, you'll no doubt notice some visual differences. This is not unexpected. I have noticed simple ones such as where VerticalAlignment="Top" is expected, but the platform renders as if VerticalAlignment="Stretch" in action with some MaxHeight setting.

I have also noticed  rather subtle ones with Visual States Management, whereby ContentControl only expands, but never shrinks back with rather annoying consequences if you want to do such things such as visual toggling of PageHeader title and AppBarButtons. This PR solves the former problem with PageHeader.

It sets the page container frame VerticalAlignment to Top (default is Stretch) and as such **breaks nothing at all** since we always want the PageHeader to be aligned with the Top anyway. VerticalAlignment Stretch can break depending on Grid Column width setting (of \* or Auto) and container MaxHeight and I suspect this is what's at play with Build 14393.

So far I observed the above problem on 4" 14393 Mobile Emulator. See my previous posting of this issue as #1178.


dg2k:
Oops. Held for further test.

PR #1236, opened 2016-09-24
Allow NavToPageAction to Navigate even if Frame is not available.

GFlisch:
In one of my project I have extended the Hamburger menu in the Shell.xaml view by adding a sub menu level (Tapped Event). The Frame is not available in the Shell.xaml (null) and navigation using the NavToPageAction is impossible.

I have added the test on the Frame properties and if null I use the  WindowWrapper.Current().NavigationServices.FirstOrDefault() and I can still navigate.

### DIFF
--- a/Template10 (Library)/Behaviors/NavToPageAction.cs
+++ b/Template10 (Library)/Behaviors/NavToPageAction.cs
@@ -26,15 +26,7 @@ namespace Template10.Behaviors
             if (nav == null)
                 throw new NullReferenceException($"Cannot find NavigationService for {Frame.ToString()}.");
 
-            var metadataProvider = Application.Current as IXamlMetadataProvider;
-            if (metadataProvider == null)
-                return false;
-
-            var pagetype = metadataProvider.GetXamlType(Page as Type);
-            if (pagetype == null)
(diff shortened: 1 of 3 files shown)


Template10 (Library)/Behaviors/NavToPageAction.cs
```diff
@@ -26,15 +26,7 @@ public object Execute(object sender, object parameter)
...
-            var metadataProvider = Application.Current as IXamlMetadataProvider;
-            if (metadataProvider == null)
-                return false;
-
-            var pagetype = metadataProvider.GetXamlType(Page as Type);
-            if (pagetype == null)
-                throw new NullReferenceException($"Cannot find TargetPage:{Page}");
-
-            nav.Navigate(pagetype as Type, Parameter, InfoOverride);
+            nav.Navigate(Page as Type, Parameter, InfoOverride);
```
JerryNixon, line 29: If the value of Page is a string because it is a view in a remote assembly, how does "as Type" work? Could you explain why you remove this logic?

JerryNixon:
Hey @GFlisch I am still interested if you can reply?

Issue #1149, opened 2016-07-28
Cancel navigation when triggered by Hamburger menu


ghost:
I'm using the Hamburger template and I navigate to my pages through the Hamburger menu (from the Shell).

I have overrided OnNavigatingFromAsync in my corresponding ViewModel but when I set **args.Cancel = true** it does prevent the navigation but the page I was navigating to is still highlighted in the shell.

Worst, when I set a delay before **args.Cancel = true**, it does not prevent the navigation at all, even if I get the deferral and complete it after the delay.

JerryNixon:
Yeah, turns out when you introduce an await in NavingFrom() Cancel no longer works because the NavService is using events instead of rewriting the underlying nav pipeline, but we're going to fix that in the next version. That being said, the highlight thing is definitely a problem. You can work around it by calling NavMenu.Highlight() but it needs to be fixed. Consider it a bug.

Stringer-C:
I've also encountered this when giving the user a choice to leave the current page and lose changes or stay on the current page.

I can half-solve it by setting **args.Cancel = true** without awaiting if there are changes, then calling an async method to display a MessageDialog - but then if the user wishes to continue with the navigation I'm having trouble figuring out which page they were trying to navigate to (unless they were trying to go back), so that I can then use the NavigationService to complete the navigation.

Is there an easy way to reference this? I've got a temporary work-around which involves adding a RequestedPage property to the NavigationService  - then setting this in NavigateAsync(), but I was hoping there might be a cleaner way.

Also, for the highlighting problem, how do I call NavMenu.Highlight() ? I can't find this anywhere.

aline-almeida:
I'm having the same problem  described by @elstringo  .
I've noticed this in NavigationService » NavigatingFromAsync  
`await dataContext.OnNavigatingFromAsync(args).ConfigureAwait(false);`

Could it be that changing the ConfigureAwait to true solves the navigation cancellation issue ?

Monsok:
@elstringo  I am having the same problem and trying your solution. How do you set the RequestedPage property? I am also interested to know how to call NavMenu.Highlight(). Cheers.

Stringer-C:
@Monsok 
Since I couldn't find another way around it, I included the T10 Library Project in my Visual Studio Solution instead of the nuget reference, then added the following property to `INavigationService`
`Type LastRequestedPageType { get; }`
[...]

PR #1237, opened 2016-09-24
Implements conditional page navigation for view-model (e.g canceling navigation for unsaved data).

dg2k:
### Background

Navigation cancellation in ViewModel is currently impossible as confirmed by @JerryNixon [here](https://github.com/Windows-XAML/Template10/issues/1149).

> Yeah, turns out when you introduce an await in NavingFrom() Cancel no longer works because the NavService is using events instead of rewriting the underlying nav pipeline, but we're going to fix that in the next version. 

Issue #1149 provides the relevant background as contributed by @bbougot, @aline-almeida, @Monsok and @elstringo. This issue is also [listed as a to-do item](https://github.com/Windows-XAML/Template10/issues/1221) for the forthcoming release 1.1.12.
### Solution

Perhaps a workaround?
- This PR adds a bool method **PageCanNavigateAwayAsync ()** to the **INavigable Interface** for the developer to _override_ and _return false_ when nav cancellation is desired.
- The default virtual implementation in **ViewModelBase** always returns true and, if not overridden, it's as if nothing has changed.
- The **FrameFacade** event  [FacadeNavigatingCancelEventHandler](https://github.com/Windows-XAML/Template10/blob/master/Template10%20%28Library%29/Services/NavigationService/FrameFacade.cs#L287) gets to see what the developer is returning via **PageCanNavigateAwayAsync ()** in ViewModel and accordingly updates its [e.Cancel](https://github.com/Windows-XAML/Template10/blob/master/Template10%20%28Library%29/Services/NavigationService/FrameFacade.cs#L307) flag. _Q.E.D._
### Usage Example

```
// ViewModel skeleton code to highlight the above

public class MyViewModel : ViewModelBase
{
    private bool PageHasDataToSave;            // set this field to true to cancel navigation

    ...

    public override async Task<bool> PageCanNavigateAwayAsync ()
    {
        return !PageHasDataToSave;
    }

        ...

    // OnNavigatingFromAsync is where most of the action happens. 
    // Assume PageHasDataToSave flag was set somewhere where data change took place.

    public override async Task OnNavigatingFromAsync(NavigatingEventArgs e)
    {
        await base.OnNavigatingFromAsync(e);
        if (PageHasDataToSave)
        {
            // Message.Show is just a helper method built around MessageDialog class [1], so just take the concept here...
            // Message.Show returns: true when OK button cliked; false when Cancel button clicked.
            bool OkToDiscard = await Message.Show(ResourceLoader.GetForViewIndependentUse().GetString("ConfirmDataChangeDiscard"), MessageDialogButton.OKCancel);
            if (!OkToDiscard)
            {
                e.Cancel = true;
                return;
            }

           // OK to discard, so negate the flag to let subsequent nav to proceed as normal.
            PageHasDataToSave = false; 
        }

        // If proceeded here, it means user has opted to discard unsaved data, so let's navigate ...
        // I wonder if the following lines could be simplified??

        if (e.NavigationMode == NavigationMode.Back)
        {
            if (BootStrapper.Current.NavigationService.CanGoBack) BootStrapper.Current.NavigationService.GoBack();
        }
        else if (e.NavigationMode == NavigationMode.Forward)
        {
            if (BootStrapper.Current.NavigationService.CanGoForward) BootStrapper.Current.NavigationService.GoForward();
        }
        else if (e.NavigationMode == NavigationMode.Refresh)
        {
            BootStrapper.Current.NavigationService.Refresh();
        }
        else if (e.NavigationMode == NavigationMode.New)
        {
            BootStrapper.Current.NavigationService.Navigate(e.TargetPageType, e.TargetPageParameter);
        }
    }
}
```

---

[[1](https://msdn.microsoft.com/library/windows/apps/br208674)]

### DIFF
--- a/Template10 (Library)/Mvvm/ViewModelBase.cs
+++ b/Template10 (Library)/Mvvm/ViewModelBase.cs
@@ -34,7 +34,6 @@ namespace Template10.Mvvm
         [JsonIgnore]
         public virtual IStateItems SessionState { get; set; }
 
-        [JsonIgnore]
         public virtual Task<bool> PageCanNavigateAwayAsync ()
         {
             return Task.FromResult(true);
--- a/Template10 (Library)/Mvvm/ViewModelBase.cs
+++ b/Template10 (Library)/Mvvm/ViewModelBase.cs
@@ -33,5 +33,10 @@ namespace Template10.Mvvm
 
         [JsonIgnore]
         public virtual IStateItems SessionState { get; set; }
+
+        public virtual Task<bool> PageCanNavigateAway ()
+        {
+            return Task.FromResult(true);
+        }
     }
 }
--- a/Template10 (Library)/Mvvm/ViewModelBase.cs
+++ b/Template10 (Library)/Mvvm/ViewModelBase.cs
@@ -34,6 +34,7 @@ namespace Template10.Mvvm
         [JsonIgnore]
         public virtual IStateItems SessionState { get; set; }
 
+        [JsonIgnore]
         public virtual Task<bool> PageCanNavigateAwayAsync ()
         {
             return Task.FromResult(true);
--- a/Template10 (Library)/Mvvm/ViewModelBase.cs
+++ b/Template10 (Library)/Mvvm/ViewModelBase.cs
@@ -34,7 +34,7 @@ namespace Template10.Mvvm
         [JsonIgnore]
         public virtual IStateItems SessionState { get; set; }
 
-        public virtual Task<bool> PageCanNavigateAway ()
+        public virtual Task<bool> PageCanNavigateAwayAsync ()
         {
             return Task.FromResult(true);
         }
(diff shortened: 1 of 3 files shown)


JerryNixon:
Wow. This is nice. I was wondering. What if we simply changed it so that OnNavigatingFrom() was no longer called by the event but instead injected as a simple call in Navigate() so that we could react to the output parameter?

JerryNixon:
The reason I am considering a different approach than yours is because I absolutely hate introducing any new signature. It should just work the way it is working, right? Adding a shim is just another thing we know we will eventually take abck away - confusing our developers if we do.

dg2k:
@JerryNixon - sorry I was away for a few days and didn't respond.

Can you please elaborate on the alternative so that I can follow you clearly?
I can understand the point that my workaround is a shim to maintain compatibility but at what point would developers get confused. If one doesn't use the proposed nav cancellation, they won't  even notice the change is there. To use nav cancellation, on the other hand, an overrideable method should be implemented. As things stand, with OnNavigatingFrom being async, I am not getting brain waves as to how the intent of ViewModel is communicated to an impending nav. Or, are you suggesting a sync nav when saying _injected as a simple call in Navigate()_ ?

JerryNixon:
https://github.com/Windows-XAML/Template10/issues/1261

JerryNixon:
Would you start to look over the 10.2 NavigationService with me? It's very early on.

PR #1269, opened 2016-10-11
Fix HamburgerMenu header visibility

schuettecarsten:
When FullScreen mode is active, the Header control must also be set to Visible=Collapsed because even it has opacity of 0, it still interacts with controls that might be behind the invisible header.


JerryNixon:
This _just_ missed v12

PR #1290, opened 2016-10-27
Added command to generate a View with its ViewModel and an optional Model

rpedretti:
I don't know about you people, but I got tired of always having to create these files one by one. So I made this command to generate all these files at once. It's my first VSIX so it will be probably messy for you (critics are welcome).

It basicaly adds a command at the context menu for a given project that pops a [dialog](https://1drv.ms/i/s!Arstk3P49tminLMGfPUDUwBG41-rag) which prompts the user for the page name, if it should generate a model and if the user wants the files to be opened once created.

I used T4 templates to generate the files based on the default structure of T10 [Project name] / [View | ViewModel | Model] and with the namespace of the project. It's also ready for localization.

I hope you find it usefull!


JerryNixon:
I think we can do this more simply with an Item Template. I tell you what @RafaelPlantard, I am impressed that you figured out the VSIX system since it is so poorly documented and works in such a cryptic way. What you might research is if you can add a command that can interact with an Item Template. Nothing wrong with right-clicking a Views folder and "Add View". The thing is, I am about to introduce a large list of Item Templates that accomplish different tasks. This will allow developers to add views that do things like add Map, add Inking, add data binding, add Image gallery, and stuff like that. I want to ensure we go down THAT bath before we worry with the VSIX. And between you and me, a VSIX package is annoyingly difficult to maintain. Please don't misread this as not being appreciative for your work. It's really great. Maybe you can be part of the team to build out item templates? I would welcome the help of developers who know how to code.

rpedretti:
I'll be glad to help! Just fill me in and let the party begin!

About the VSIX, my idea wes to give the user the minimum amount of options. Sometimes it's difficult to choose between a user control or a templated control, so just creating a "page" did the job. But yes, I spent a couple days figuring out the constants and functions. A bit of a pain but the needs of the many outweigh the need of the few, right?

PR #1347, opened 2016-12-14
HamburgerMenu - HighLightCorrectButton

GFlisch:
I have created a new one based on the latest master branch.

In the HamburgerMenu when a navigation starts, a search is started to find if a button exists to high light it.

The search criteria is first based on the type of the page and after on the PageParameter.
The select is based on 2 criterias: null or PageParameter equal to pageParam.

The error consists in the fact that if 2 buttons exist (one null and one with a parameter) the null is taken.

Introducing a OrderByDescending will force to select first the button matching the pageParam critieria.

            // add parameter match
            buttons = buttons.Where(x => Equals(x.HamburgerButtonInfo.PageParameter, null) || Equals(x.HamburgerButtonInfo.PageParameter, pageParam));
            var button = buttons.OrderByDescending(x => x.HamburgerButtonInfo.PageParameter)
                                .Select(x => x.HamburgerButtonInfo).FirstOrDefault();
            Selected = button;

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -195,3 +195,4 @@ Templates (Project)/Blank/null
 /packages/Microsoft.VSSDK.BuildTools.14.1.24720
 
 *.zip
+/Template10 (Installer)/Template10.Installer/ProjectTemplates/Windows/Universal/Template10/Temp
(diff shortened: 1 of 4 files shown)


dg2k:
Could this be a source of something strange I am observing with the latest nav?
I haven't done a detailed investigation but I made some observation that MainPage.xaml is loaded twice on startup. I somehow assumed this has to do with the binding update (to counter the async nav wrt OneTime binding) but wasn't sure if this is limted to OnNavigatedToAsync or if it affects the Loaded event as well (which I noted). Just in case the two are related ...

JerryNixon:
Looks great. You will need to update your PR so I can accept it.

JerryNixon (commented):
Can you update?

PR #1359, opened 2016-12-25
add the AwaitableDelegateCommand code snippet.

yanxiaodi:
I think it's more helpful to add this snippet when I use the AwaitableDelegateCommand.


JerryNixon:
You know. This is not my favorite feature of T10 and I am more inclined to delete it than fix it. The implementation is quite confusing and I think zero developers are using it; other than you. :)

PR #1364, opened 2016-12-29
Makes GoBack() and GoForWard() FrameFacade methods public.

dg2k:
Changes the access modifier from internal to public for GoBack() and GoForWard() FrameFacade methods.


JerryNixon:
What's wrong with the navigation service methods?

JerryNixon:
Only the navigation service providers View model interaction. A developer calling the frame façade methods might be confused, right?

dg2k:
My bad! When we made ForwardStack and BackStack properties public, I thought we overlooked these closely related methods -- of course already exposed via the NavigationService.

@JerryNixon - this raises an interesting question. Why not provide ForwardStack and BackStack properties on NavigationService as well rather than get to them via FrameFacade?

Issue #1370, opened 2017-01-03
Taming the Asynchronous Navigation (Cont'd)


dg2k:
Sorry, got interrupted with what I was saying in #1369, and here I continue.

- In [UpdateSelectedAsync()](https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Controls/HamburgerMenu.xaml.cs#L533) method, the call that triggers navigation cancellation dialog is [this one](https://github.com/Windows-XAML/Template10/blob/master/Template10%20(Library)/Controls/HamburgerMenu.xaml.cs#L560) with the code line `if (await NavigationService.NavigateAsync...`. [Please note: these links may not work if the code has changed].
- Now, we can safely postpone the `previous.IsChecked = false;` and `previous.RaiseUnselected();` lines after the cancellation dialog, saving us the hassle of undoing for the page we attempted to navigate from.
- The last **else** block of the **if** condition catches a cancelled navigation. There we reverse what we did to the new page whereby `value.IsChecked = false;` and `value.RaiseUnselected();` do the trick.
- But wait -- remember the new page nav started with **Selected** DP setting to HamburgerMenu's **HamburgerButtonInfo** , so **Selected** needs to be restored to the current page we're staying with after nav cancellation. This one is tricky: setting `Selected = previous;` triggers a recursive call, but luckily all that is required setting is **Selected** without the navigation that would follow.
- And a flag for nav cancellation comes to a rescue; I called this flag **IsCanceledNavigation** and is set before calling  `Selected = previous;` and reset afterwards thereby avoiding the unnecessary nav elsewhere.
- The proposed solution also eliminates the problem of double-hit of a page which has been reported by some devs over the recent days.
- **Disclaimer** - still under test and not yet fool-proof!
A PR is coming up shortly [okay, now posted as #1371], and what I've described is working with my app as expected with more testing going on. Hope @JerryNixon will approve this.

Issue #1369, opened 2017-01-03
Taming the Asynchronous Navigation


dg2k:
Happy New Year to all at **Template10** and a happy 2017 coding!

I'm sure you'll agree that **async navigation** is perhaps the most challenging part of T10, particularly where nav needs to be canceled as in situations when a page to be navigated from has unsaved data. The implementation of a fully working **OnNavigatingFromAsync** is still work in progress and here are my observation of the rough edges that need to be smoothed out.

- With HamburgerMenu, all page navigations commence with the **NavButtonChecked()** event handler in **HamburgerMenu.xaml.cs**. This sets the **IsChecked** property for the menu and sets `Selected = button.HamburgerButtonInfo;` which initiates navigation to a point where navigation cancellation dialog is presented to the user. (For this discussion, leave aside the back and forward navigation which is less problematic).

- By the time the user chooses to cancel the navigation (in page's **OnNavigatingFromAsync** ), the page to be navigated to is already highlighted as checked (as a menu item) and **Selected** as the current menu page. Yes, navigation is cancelled but working with the HamburgerMenu requires more to undo and get everything back to the state before the navigation was attempted ('cos we canceled it!).

- The above is not a trivial task to solve. Ideally we would like to postpone setting a new page to be navigated to and relinquishing the old page until navigation cancellation decision is reached. But we have no choice with the former (i.e., new page nav) because navigation has to happen for it to be cancelled!  For the latter, *can we change* navigation updating logic such that we postpone relinquishing the current page until navigation cancellation decision is reached? Close inspection of the **UpdateSelectedAsync()** method in **HamburgerMenu.xaml.cs** shows that there is a scope for this.

To continue ... [as #1370]

PR #1371, opened 2017-01-03
Tames the async nav as per discussion in issues #1369 and #1370.

dg2k:
Okay, strange term *tame* but that's what it is ;-)
See #1369 and #1370.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
@@ -504,9 +504,9 @@ namespace Template10.Controls
 
         private bool IsCanceledNavigation;
 
-        private async Task UpdateSelectedAsync(HamburgerButtonInfo previous, HamburgerButtonInfo value)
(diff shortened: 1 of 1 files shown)


Template10 (Library)/Controls/HamburgerMenu.xaml.cs
```diff
@@ -504,9 +504,9 @@ public void UpdatePaneMarginToShowHamburgerButton()
 
         private bool IsCanceledNavigation;
 
-        private async Task UpdateSelectedAsync(HamburgerButtonInfo previous, HamburgerButtonInfo value)
+                private async Task UpdateSelectedAsync(HamburgerButtonInfo previous, HamburgerButtonInfo current)
```
schuettecarsten: Check for spaces/tabs

dg2k:
I think that was due to the Online Editor. I used it to save me time (for minor correction) and suspect it may not be perfect? [On second thought, it could be that I got it wrong with copy-aste misalignment].  I will try to correct it.
Thanks for the feedback @schuettecarsten.

Correction done!

dg2k:
Okay, will steer away from the Online Editor; here is PR #1374 of the same with the white space correction.

Issue #1376, opened 2017-01-10
HamburgerMenu.HighlightCurrentButton causes Navigation events on "Back"


schuettecarsten:
HamburgerMenu.HighlightCurrentButton causes useless navigation events when "Back" button is used.

schuettecarsten:
There was a bug in my chanes, please also merge #1382

PR #1378, opened 2017-01-10
Prevent unneccessary Navigation on "Back" caused by HighlightCorrectButton

schuettecarsten:
HighlightCorrectButton updates the Selected hamburger button, the property itself triggers methods and navigation requests. This causes some useless navigation events when "Back" button is used or on application startup. This PR will fix this, #1376


schuettecarsten:
The UpdateInternalBindingValues call is not absolutely neccessary, but during my tests, I have had a serious problem where one button was suddenly disabled. This is why I decided to update the internal state whenever the state is updated.

PR #1382, opened 2017-01-14
HighlightCorrectButton did not unselect button when navigating to unknown View

schuettecarsten:
This PR fixes a regression in my latest changes. HighlightCorrectButton did not unselect the latest Button if a View is shown that is not known in the HamburgerMenu.

### DIFF
--- a/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
+++ b/Template10 (Library)/Controls/HamburgerMenu.xaml.cs
@@ -394,6 +394,8 @@ namespace Template10.Controls
             _isHighlightCorrectButtonRunning = true;
             try
             {
+                HamburgerButtonInfo newButton = null;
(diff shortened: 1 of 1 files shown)


JerryNixon:
This `UpdateInternalBindingValues` is really a good addition. 
Can you refresh from master so I can merge this?

schuettecarsten:
Updated. Please note that the main idea in my changes is that the button is only touched when the selection really changes. So I moved some code into the ReferenceEquals-if-block. But you're right, it should not be a problem to update the internal binding values every time, but if you call it for the new button, you should also call it for the old button.

schuettecarsten:
My latest commit is neccessary to make this PR work after the changes in PR #1388. So please make sure to merge #1388 before this PR or revert commit 249f368.

PR #1384, opened 2017-01-15
Adds an if condition not accounted for, namely (viewmodel != null) 

dg2k:
The if  condition in question is based on view pages's DataContext (usually ViewModel) but what if it's a different kind of DataContext? You may be testing something different but a good idea to account for DataContext that is not for a ViewModel. I discovered this in the process of changing a view page to view-model and the page had a DataContext set in codebehind - took me a while to figure out the strange behavior as T10's navigation returns true on exhausting the if condition test.

Better return false same as the `else if (viewmodel == null)`.  Is this related to #1380? May be the same from a different angle?


dg2k:
Sorry, I have to close this and re-open again. Untidy commit info.

PR #1385, opened 2017-01-15
Adds an 'if condition' unaccounted for, namely (viewmodel != null).

dg2k:
The if condition in question is based on view pages's DataContext (usually ViewModel) but what if it's a different kind of DataContext? You may be testing something different but a good idea to account for a DataContext that is not for a ViewModel.

I discovered this in the process of adding a viewmodel and the view page had a DataContext of different type set in code-behind (and a while in existence) - took me sometime to figure out a strange behavior as T10's navigation returns true on exhausting the if condition test.

Is this related to #1380? I think it's and same problem from a different angle?

### DIFF
--- a/Template10 (Library)/Services/NavigationService/NavigationLogic.cs
+++ b/Template10 (Library)/Services/NavigationService/NavigationLogic.cs
@@ -171,6 +171,10 @@ namespace Template10.Services.NavigationService
                 var vm = viewmodel as Portable.IConfirmNavigationAsync;
                 return !await vm?.CanNavigateAsync(parameters);
             }
+            else if (viewmodel != null)
+            {
+                return false;
+            }
             else
             {
                 return true;


schuettecarsten:
This change might be breaking: Of course, you _can_ navigate to a view/viewmodel that does not have any of the interfaces implemented, why not? But you will not get any methods called.

dg2k:
I suspected of the breaking change but good to know *the unknown*,  in this case the *unaccounted-for condition*. Then we can work out a better solution.

Template10 (Library)/Services/NavigationService/NavigationLogic.cs
```diff
@@ -171,6 +171,10 @@ public void SetupViewModel(INavigationService service, ITemplate10ViewModel view
                 var vm = viewmodel as Portable.IConfirmNavigationAsync;
                 return !await vm?.CanNavigateAsync(parameters);
             }
+            else if (viewmodel != null)
```
JerryNixon, line 174: I don't understand why `else { return false; }` isn't all we needed. Will `else` ever be called?

dg2k: @JerryNixon But `else` is returning `true`. See the original code. We don't want `true` returned for non ViewModel datacontext.

schuettecarsten: @dg2k: But `false` will abort navigation at all. With `false`, you cannot navigate to datacontexts that do not implement any of the interfaces. You have to implement at least one of them, even if your code does not require them.

dg2k: Isn't not navigating better if the DataContext is not the expected ViewModel?
This is an edge-case I came across while changing from view page to view-model and thought a good idea to catch this out. If someone (like I did!) is messing up with a non-compliant datacontext it's better not to navigate.

What would you do with a  different type of datacontext? perhaps throw an exception? It's very rare but not impossible. From programming point of view, the if-condition exists and something needs to be done rather than leave it as unknown.

May be I didn't make my assumption clear. Navigation for any other datacontext should fail (hence not navigate) because in all likelihood it will eventually fail (but hard to track if did navigate and then fail as I witnessed with one case).

JerryNixon: Will `else` ever be called?

JerryNixon: Again, will `else` ever be called?

dg2k: Good question ...
Here is how it all started:
The **else** was returning **true** before my PR hence why I kept at that by excluding the situation I encountered. I thought it was for a reason **else** was returning **true**, so I didn't look at it very closely beyond the case of some random DataContext that was not ViewModel. If returning **true** was not intended in the first place, then poblem solved.

If the following is true, we can end the condition block with else returning false.

> viewmodel is some non-null DataContext and not of the Type checked against in the preceding tests.

I hit the **else** that was returning **true** with a view page DataContext and that triggered the PR.

dg2k: Sorry, forgot hitting the submit review button and only noticed now after days...

PR #1390, opened 2017-01-18
Sample SettingsPage.OnNavigatedTo() defends against calling ToString …

angrah:
…on a null e.Parameters, but then calls ToString() on the result of Deserialize(null) which is also null.

A recent change removed the parameter from the navigate in Shell.xaml

### DIFF
--- a/Templates (Project)/Hamburger/Views/SettingsPage.xaml.cs
+++ b/Templates (Project)/Hamburger/Views/SettingsPage.xaml.cs
@@ -17,8 +17,11 @@ namespace Sample.Views
 
         protected override void OnNavigatedTo(NavigationEventArgs e)
         {
-            var index = int.Parse(_SerializationService.Deserialize(e.Parameter?.ToString()).ToString());
-            MyPivot.SelectedIndex = index;
+            if (e.Parameter != null)
+            {
+                var index = int.Parse(_SerializationService.Deserialize(e.Parameter.ToString()).ToString());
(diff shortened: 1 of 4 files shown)


Templates (Project)/Hamburger/Views/SettingsPage.xaml.cs
```diff
@@ -17,8 +17,11 @@ public SettingsPage()
 
         protected override void OnNavigatedTo(NavigationEventArgs e)
         {
-            var index = int.Parse(_SerializationService.Deserialize(e.Parameter?.ToString()).ToString());
-            MyPivot.SelectedIndex = index;
+            if (e.Parameter != null)
+            {
+                var index = int.Parse(_SerializationService.Deserialize(e.Parameter.ToString()).ToString());
```
JerryNixon: This does not, however, defend against Deserialize returning null.

angrah: That's true, but it didn't before this change either.

angrah: The JsonSerializationService doesn't return null unless you feed it null or it got null from Json.DeserializeObject(). Better to be safe I guess you're right. I will add the extra elvis and update this pull request.

angrah:
Some of these changes already snuck in with my other pull request. Still getting the hang of this git thing.

PR #1405, opened 2017-02-13
Sample Search: Fixed UI thread blocking

snrnats:
Fix for a sample app.


schuettecarsten:
The fix is invalid. Never use DispatchAsync without await or Task.Run because this might cause that the task only runs to the first await and then stops. I think the code should be like this:

```
            Task.Run(() => WindowWrapper.Current()
                .Dispatcher.DispatchAsync(() => Shell.ShowBusy(false, string.Empty), 5000));
```

snrnats:
@schuettecarsten 
Could you please give more details about the case that will cause the task only runs to the first await.
Also there is one more usage of DispatchAsync without await or Task.Run in Samples\MultipleViews\ViewModels\MainPageViewModel.cs:51

`wrapper.Dispatcher.DispatchAsync(() => Messages.Add(message.Content));`

PR #1409, opened 2017-02-25
Use "await DispatchAsync" instead of "await Dispatch" when returning a task

schuettecarsten:
Use "await DispatchAsync" instead of "await Dispatch" when the inner method is returning a task. Because you want to await the inner task in the dispatcher and the outer task is awaiting the dispatcher. If you use Dispatch, when you will return the task and await it from outside. This is wrong.


JerryNixon:
Yes, all good changes

JerryNixon:
I still want this. If you want it in 10.1 target that branch, otherwise, 10.2 welcomes it.

PR #1410, opened 2017-02-25
Workaround for too large State Manager setting

schuettecarsten:
Workaround: If state manager setting cannot be written on suspend, clear the values and exit the App to force a clean startup next time.

### DIFF
--- a/Template10 (Library)/Common/Bootstrapper/BootStrapper.cs
+++ b/Template10 (Library)/Common/Bootstrapper/BootStrapper.cs
@@ -699,7 +699,23 @@ namespace Template10.Common
                     {
                         // individual frame-level
                         await nav.SuspendingAsync();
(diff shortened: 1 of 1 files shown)


Template10 (Library)/Common/Bootstrapper/BootStrapper.cs
```diff
@@ -699,7 +699,24 @@ private void SetupLifecycleListeners()
...
                         // individual frame-level
                         await nav.SuspendingAsync();
-                        if (AutoSuspendAllFrames) await nav.SaveAsync();
+                        if (AutoSuspendAllFrames)
+                        {
+                            try
+                            {
+                                await nav.SaveAsync();
+                            }
+                            catch
```
JerryNixon, line 708: Don't you want to call SaveAsync() again?

PR #1424, opened 2017-03-21
Fixed issue that caused changes in a shared brush

snrnats:
It changes opacity of a brush that can also be used by other controls.

e.g.
`<Style TargetType="controls:HamburgerMenu">
<Setter Property="HamburgerBackground" Value="{ThemeResource SystemControlBackgroundAltHighBrush}" />
    </Style>`
causes changes everywhere we use `SystemControlBackgroundAltHighBrush`


JerryNixon:
Fix this! I want it!

snrnats:
Updated

PR #1455, opened 2017-05-02
Cleaned, Washed, and Wax...?. And all compile excl. VSIX. [later: and wiped the tabs]

dg2k:
@JerryNixon - Just a little help with cleaning the references as I noticed some loose ends with the projects under VisualStudio folder (aka Template projects).

- VS2017's Power Command (Remove and Sort Usings) applied to the whole solution; That's one powerful vacuum cleaner, courtesy of VS2017. (**Side Note:** On one occasion I noticed VS2017 slurped all the references in ALL AssemblyInfo.cs files, BUT couldn't reproduce this phenomenon afterwards. Strange, so could be a random bug??. The morale ... make sure you make a backup before you unleash this Power Command.)

- No code touched, except a few in Template projects mentioned below.

- Template projects updated with NuGet Package Manager

- A stray empty project removed. I note that this has already moved to Template10 folder as **StateService** project.

- Minor corrections in Template project settings service.

- Glad to see that the **Default** navigation service property is now assigned with a value. I had a PR ready for it but ...

- All compile with VSIX project excluded.


tibitoth:
BTW Why do we need to remove and sort usings, ecspecially the common System.* ones?

dg2k:
True, I wish System is left behind but that's how VS2017 (or even versions before do).
Don't know is I overlooked an option on this.

There is an option to put System at the top regrardless of alphabetical order.

Sorry I'm closing this too (with a replacement of course) as I noticed something silly ...

PR #1498, opened 2017-06-08
KeyboardService / BackService / ModalDialog

mvermef:
KeyboardService is complete **not** published but other files directly dependent are referencing via local nuget repository on my dev machine it does work.  Only issue was coming up with a Sample still thinking

BackService really nothing much to it, no sample to really demonstrate.  Since its still a dependent service, navigation has the best example of this service.

Modal Dialog: complete with package unpublished.  Sample works. Based on sample for Search in original samples.


mvermef:
now that I think about it I should probably break theses up.

PR #1502, opened 2017-06-11
Adds nuget ignorables (.targets and .props).

dg2k:


dg2k:
oops, was was not supposed to be closed!

Issue #300, opened 2015-11-04
NavigationService in ViewModel - Object reference not set to an instance of an object


ghost:
I'm trying to navigate from my ViewModel to a page.
I followed this tut: https://marcominerva.wordpress.com/2015/10/07/navigationservice-with-mvvm-light-and-template-10/

But doing : NavigationService.Navigate(Pages.MainPage); from my ViewModel doesn't work, NavigationService is null (my ViewModel inherits ViewModelBase)

Any thought?

MichaelPetrinolis:
Have you tried the sample from the repo?

ghost:
Yes but the sample doesn't call NavigationService.Navigate from the viewmodel.
Anyway, I solved my problem with a simple workaround: I injected the NavigationService instance from the Bootstrap class in my DI so that I can use it in my ViewModel.

In the App class:

``` csharp
        public override Task OnInitializeAsync(IActivatedEventArgs args)
        {
            DispatcherHelper.Initialize();
            SimpleIoc.Default.Register(() => NavigationService);

            return base.OnInitializeAsync(args);
        }
```

In my ViewModel:

``` csharp
        public StartTripViewModel(INavigationService navigationService)
        {
            NavigationService = navigationService;
            StartTripCommand = new RelayCommand(async () =>
            {
                var contacts = await GetContacts();
                NavigationService.Navigate(Types.Pages.Contact, contacts);
            });
        }
```

marcominerva:
Hi bbougot!

I actually use NavigationService from ViewModel without problems. If you send me an email from https://marcominerva.wordpress.com/contacts/, I would be happy to help you identifying the problem.

JerryNixon:
@bbougot Your solution is possibly problematic because the bootstrapper will inject the navservice for you, and the version of the navservice it injects will be the correct thread and frame-ref. Your approach is okay for you, but is not generally a good practice.

JerryNixon:
The real question is why the navservice would EVER be null. It's important that you repro this for us so we can correct this for other cases like yours.

marcominerva:
I'm currenlty in touch with bbgout to figure out the problem. As soon as we found the issue, I'll update this post with the details.

ghost:
I have the same issue. Was there ever a resolution to this? My ViewModelBase implements INavigable, and my view models all inherit from ViewModelBase. However, when I call this.NaviationService.Navigate(typeof(MyPage), null); I see that this.NavigationService is null.

[...]

PR #1527, opened 2017-08-15
OnNavigatedToAsync() not called when overriding ResolveForPage()

dackjaniels2001:
When overriding ResolveForPage() in App.xaml.cs in order to instantiate viewmodels in an MVVM architecture, the OnNavigatedToAsync() method is never called and the resulting viewmodel has a null NavigationService.

Assigning the newViewModel variable in NavigationService.NavigationOrchestratorAsync() means that the following call to NavToAsync() does not short circuit due to viewmodel being null. This in turn means the viewmodels OnNavigatedToAsync method gets called and the viewmodels NavigationService property gets set correctly.

### DIFF
--- a/Template10 (Library)/Services/NavigationService/NavigationService.cs
+++ b/Template10 (Library)/Services/NavigationService/NavigationService.cs
@@ -202,7 +202,7 @@ namespace Template10.Services.NavigationService
             // call newViewModel.ResolveForPage()
             if (newViewModel == null)
             {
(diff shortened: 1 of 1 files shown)


dackjaniels2001:
I believe the same problem is described in [this](https://github.com/Windows-XAML/Template10/issues/300#issuecomment-154587972) (rather old) GitHub issue.

JerryNixon:
Good catch

Issue #1346, opened 2016-12-14
Introducing SecretService


JerryNixon:
I have documented it here: https://github.com/Windows-XAML/Template10/wiki/Services#secretservice

dg2k:
@JerryNixon  -- I use PasswordVault() in the Windows.Security.Credentials ns a lot and this is a wonderful addition to T10 for security conscious folks. Thank you!
... sure the name SercretService doesn't cause some copyright infringement? Only kidding ;-)

mvermef:
or the NSAs ears are twitching.... 😮

tibitoth:
Where is this service's code?

JerryNixon:
Hither https://github.com/Windows-XAML/Template10/tree/master/Template10%20(Library)/Services/SecretService

JerryNixon:
There is a bug in the current version if you attempt to store string.Empty as a secret. I have fixed this in a branch and will merge it soon. Probably after my vacation.

michaelhutchful:
I can't find the SecretService in the Nuget package. How can I use it?

yanxiaodi:
Very cool! This function is helpful to store some sensitive values like token, etc.

mikea314:
The SecretHelper class doesn't persist modified secrets due to a missing ' _vault.Add(credential) ' line:

    public void WriteSecret(string container, string key, string secret)
        {
            if (_vault.RetrieveAll().Any(x => x.Resource == container && x.UserName == key))
            {
                var credential = _vault.Retrieve(container, key);
                credential.RetrievePassword();
                credential.Password = secret;  
                _vault.Add(credential); <==== Required to persist change
            }
            else
            {
                var credential = new PasswordCredential(container, key, secret);
                _vault.Add(credential);
            }
        }

dg2k:
You're correct. I thought I made a PR on this a while back and somehow fell thru the crack (see bottom of #1503 and can't figure out what really happened then). There was a lot changing at the time.

PR #1531, opened 2017-08-20
Adds SecretService Demo with some fixes.

dg2k:
Adds SecretService Demo with some fixes  to **SercretService**.
- Secret service credential is now persisted -- see comments toward end of #1346 page.
- Convenience method `IsSecretExistsForKey()` is added to **SecretHelper.cs**
- Nuget package as **Template10.Services.SecretService v1.0.2** already live.


JerryNixon:
Want to change to `Template10.Services.Secrets` and include `SecretService` but also include `CrytoService` in the same namespace. Take a look at what we did with `Template10.Services.Serialization` if you want to know what I mean.

dg2k:
Oh! somehow missed this message. Will have a look at this.

PR #1539, opened 2017-08-22
10.2 - DI container accepts keyed registration/resolve

rpedretti:
Added new methods to the IContainerService interface and added a demo app with the using the new interface


rpedretti:
I also had to change the navigation service a bit so on Back navigation the ViewModel will also be resolved. I tested With Unity and it works with and without navigation cache, With navigation cache the resolved ViewModel is the same instance as the page DataContext. With disabled cache the DI container returns a new instance.

JerryNixon:
Closed

PR #1546, opened 2017-08-23
Example of a ViewModel resolver

rpedretti:


JerryNixon:
The app can register view-models in `OnInitializeAsync()` and should inherit from `MvvmLightBootStrapper` and not `BootStrapperBase`. Ack! Obviously an unclear API there. Quite different from T10.1. I need to build out Minimal, maybe with a "Typical" template. I think maybe a `DefaultBootStrapper` might also make sense, later. Hmm. Thoughts?

rpedretti:
I agree on the ```OnInitializeAsync()```. But inheriting from ```MvvmLightBootStrapper``` won't let me choose my DI container. I've build the ```DefaultBootStrapper``` and it Works, but maybe it's not the ideal approach, at least not with this name. I've made the dependency registration a three step action

```
public BootStrapperBase(IContainerService containerService)
{
    // start
    Current = this;
    Services.Container.ContainerService.Default = containerService;
    RegisterMessagingService();
    RegisterInternalDependencies();
    RegisterDependencies();
    .......

public abstract void RegisterMessagingService();
public abstract void RegisterInternalDependencies();
public virtual void RegisterDependencies() { }

....
```


At the abstract ```DefaultBootStrapper``` i've implemented ```RegisterInternalDependencies()``` as it was on the MvvmLight ```RegisterDependencies()```, except the messaging system, wich need to be registered like:
```
public override void RegisterMessagingService()
{
    ContainerService.Register<IMessengerService, MvvmLightMessengerService>();
}
```

I made like this so the ```DefaultBootStrapper```  won't have any reference to the actual messaing implementation.

JerryNixon:
You're right. Let me resolve this somehow.

rpedretti:
Ok. Tell me if there is something I can help with.

JerryNixon:
I will build this for you. In the new Template.Advanced.

rpedretti:
Ok! I don't know if you faced a navigation problem while navigating back, but the current NavigationService won't resolve ViewModels and there will be NullReferences everywhere, unless pages are cached.

rpedretti:
Just saw the updated code at master, ignore me

PR #1614, opened 2018-02-05
Repaired sources uri for control themes

jp-weber:
Controls could not be load because the sources cannot be found => fix sources.

### DIFF
--- a/Template10.Controls.Library/Themes/Generic.xaml
+++ b/Template10.Controls.Library/Themes/Generic.xaml
@@ -4,9 +4,9 @@
     xmlns:controls="using:Template10.Controls">
     
     <ResourceDictionary.MergedDictionaries>
-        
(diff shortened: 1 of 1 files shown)


JerryNixon:
This is a platform change, and I greatly appreciate the update.

JerryNixon:
Can you update this in the 13b branch instead?

JerryNixon:
I will also update these in master, yes.

jp-weber:
Isn't the source correct? 
https://github.com/Windows-XAML/Template10/blob/version_1.1.13p/Template10%20(Library)/Template10%20(Library).csproj#L11
The assembly name is Template10Library:
https://github.com/Windows-XAML/Template10/blob/version_1.1.13p/Template10%20(Library)/Themes/Generic.xaml

PR #1633, opened 2018-03-21
Fix retrieving navigation parameter by key.

ShawnTheBeachy:
I was using:

```c#
NavViewProps.SetNavigationUri(myNavItem, PathBuilder.Create("MusicPage", ("sectionKey", section.Key)).ToString());
```

And then trying to retrieve it:

```c#
_sectionKey = parameters.GetValue<string>("sectionKey");
```

which resulted in "the given key was not present in the dictionary." I think this is the correct fix for this, but I'm happy to modify it if not.


ShawnTheBeachy:
Just realized some other stuff that I was working on sneaked in here. Closing in favor of #1640.

PR #1647, opened 2018-07-04
Fix navigation parameters on GoBackAsync

ShawnTheBeachy:
Previously, when navigation parameters were handed into `GoBackAsync`, the navigation service ignored them and used `default(INavigationParameters)`, which resulted in the navigation service using whatever parameters had originally been handed to the page. This change respects the provided parameters.


jp-weber:
I implement your pull requests in my repo and it works very well. The navigation works for me now.

PR #1668, opened 2018-09-23
fix navigation path order (GetNavigationPath)

jp-weber:
Solves a problem that the navigation path has a wrong order.
Sample navigation:
/MainPage
SettingPage?item=1
SettingPage?item=2

Currently it becomes /SettingPage?item=2/MainPage/SettingPage?item=1
and not /MainPage/SettingPage?item=1/SettingPage?item=2


JerryNixon (approved):
Great catch; subtle problem; thanks

PR #1669, opened 2018-09-23
Extended TryFindItem() with parameter and fix GoForwardAsync()

jp-weber:
… to distinguish between multiple settings pages. I also updated NETCore libraries.

In the NavigationService I fixed GoForwardAsync(), so the correct stack entry is taken now.


JerryNixon:
Next time, if you are going to update any library, do that as a single PR.

jp-weber:
Okay, I'll do it.

PR #1670, opened 2018-09-29
 Allow setting the NavViewEx selected item without performing navigation

jp-weber:
This is sometimes necessary to properly set the initially selected item.

/by TastesLikeTurkey


jp-weber:
If we merge this commit, do we need the [Initialize ](https://github.com/Windows-XAML/Template10/blob/master/Source/Template10.Extras.16299/Controls/NavViewEx.cs#L163)function and the [IsStartPageProperty](https://github.com/Windows-XAML/Template10/blob/master/Source/Template10.Extras.16299/Controls/NavViewProps.cs#L23)?

PR #1732, opened 2020-09-05
DialogService update

jp-weber:
Extends the dialog service with the option to close dialogs by token and to query whether dialogs through the service are currently active.


Source/Template10.Services.Dialog/Abstractions/IDialogService.cs
```diff
@@ -12,6 +12,10 @@ public interface IDialogService
 
         Task<ContentDialogResult> ShowAsync(ContentDialog dialog, TimeSpan? timeout = null, CancellationToken? token = null);
 
+        void CancelDialogs();
```
JerryNixon, line 15: Should this be `CloseDialogs()`?

jp-weber: Yes you are right, that is better.

jp-weber:
@JerryNixon 
Coming soon I want to extend (or move to Template10.Core) the service so that the ViewModel can also be loaded for the ContentDialog.

I like the implementation in the Prism project very much:
https://github.com/PrismLibrary/Prism/blob/master/src/Uno/Prism.Uno/Services/Dialogs/DialogService.cs#L55

